PR #19, opened 2023-01-03
Patches 1.41

RedFox20:
Fix: handle AudioDevice failure
Fix: handle Graphics Device reset by unloading and reloading universe models
Fix: handle shield graphics rendering better, including automatic effect reloading
Debug: improved error logging and new CLI option --console
Fix: AutoPatcher runs Cleanup tasks before restarting the game
Fix: ResearchScreen crashes due to bad game data

### DIFF
--- a/Ship_Game/GameScreens/Universe/ResearchScreenNew.cs
+++ b/Ship_Game/GameScreens/Universe/ResearchScreenNew.cs
@@ -6,7 +6,6 @@ using SDGraphics;
 using SDUtils;
 using Vector2 = SDGraphics.Vector2;
 using Rectangle = SDGraphics.Rectangle;
-using System.Linq;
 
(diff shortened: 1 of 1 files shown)


Ship_Game/GameScreens/Program.cs
```diff
@@ -3,187 +3,199 @@
...
+                GlobalStats.LoadModInfo(value);
+                if (!GlobalStats.HasMod)
+                    throw new Exception($"Mod {value} not found. Argument was: {arg}");
             }
-            finally
+            else if (key == "--export-textures")
             {
-                Log.FlushAllLogs();
-                Environment.Exit(exitCode);
+                GlobalStats.ExportTextures = true;
```
gkapulis, line 83: why not switch case here?

PR #63, opened 2023-04-15
Feature/mars141 remove AO and ship pool

gkapulis:


Ship_Game/Ships/IShipPool.cs
RedFox20, line 1: A bit worried that this was removed. It did serve a purpose to avoid inserting a Ship into multiple AI managed pools. If we still have AI managed ship lists, then we should keep this around, rename it / adapt it

Ship_Game/Universe/UniverseScreen/UniverseScreen.InputActions.cs
```diff
@@ -37,14 +37,6 @@ public void DoExplore(Ship s)
             s.AI.OrderExplore();
         }
 
-        public void DoDefense(Ship s)
```
RedFox20, line 40: So if all of this code was broken, what does the AI even do to protect its Empire?

Ship_Game/Ships/IShipPool.cs
gkapulis: We dont have AI managed list right now. But we have this PR to reinstate IShipPool if we ever need it.

Ship_Game/Universe/UniverseScreen/UniverseScreen.InputActions.cs
gkapulis: When the AI detects a fleet coming in to a system, It creates a defense fleet. The defense fleet can also summon bombers and troops if the AI planet was taken - it will first clear any defense in the system and then get bombers and troops, So it can retake lost assets. The AI can also cancel other fleet tasks in order to create defense tasks. It is all part of the war logic I redesigned quite some time ago.

PR #74, opened 2023-06-14
remove damage divider from explosion damage and apply 25% of the explosion damage to each quadrant.

gkapulis:


Ship_Game/Ships/Ship_ModuleGrid.cs
```diff
@@ -526,17 +526,19 @@ public ShipModule HitTestSingle(Vector2 worldHitPos, float hitRadius, bool ignor
...
-                }
 
                 if (mq.Type == DamageTransfer.Root)
                 {
-                    // explosion damage will now be whats left after root module exploded
-                    damageAmount = remainingDamage; 
+                    if (mq.Module.DamageExplosive(damageSource, ref rootDamage))
+                        return; // Root module absorbed all the explosion
+                    else
+                        damageAmount = remainingDamage = diagonalDamage = rootDamage; // This is the new damage after root module exploded
```
RedFox20, line 535: This is a bit confusing, doesn't it apply the original root damage to the damage amount?

Ship_Game/Ships/Ship_ModuleGrid.cs
```diff
@@ -510,7 +511,6 @@ public ShipModule HitTestSingle(Vector2 worldHitPos, float hitRadius, bool ignor
             // If point 1 absorbs the damage it wont spread to other points.  
             // Damage is spread from point 1 to point 3 upwards, then from point 1 to point 3 backwards.
             // Then it will start from module 2 Diagonaly and repeat the logic. 
-            // These numbers are also divider for any excess damage transfered to the next module in the generator
             // Excess damage is transferred diagonally as well.
             foreach (ModuleQuadrant mq in EnumModulesQuadrants(worldHitPos, hitRadius, !ignoreShields))
```
RedFox20, line 515: It would be much easier to understand and modify if this simply called each quadrant explicitly. You can see an example in `PowerGrid.DistributePowerFrom` which is incredibly fast for what it does. There are some special optimizations there to speed it up, for example handling of small radiuses is a special case.
```cs
public void DamageExplosive(GameObject damageSource, float damageAmount,
                            Vector2 worldHitPos, float hitRadius, bool ignoreShields)
{
    Vector2 localPos = WorldToGridLocal(worldPos);
    Point c = GridLocalToPoint(localPos);
    // ... validate inputs etc ...

    int diameter = (int)Math.Ceiling((hitRadius * 2) / 16f);
    if (diameter == 1) // 1x1 explosion, direct damage
    {
        foreach (ShipModule m in GetModulesAt(c, !ignoreShields))
            if (m.DamageExplosive(damageSource, ref damageAmount))
                return true;
    }
    else // 2x2 or bigger explosion
    {
        DamageQuadrant(damageSource, c, diameter, damageAmount * 0.25f, Quadrant.NW);
        DamageQuadrant(damageSource, c, diameter, damageAmount * 0.25f, Quadrant.NE);
        DamageQuadrant(damageSource, c, diameter, damageAmount * 0.25f, Quadrant.SE);
        DamageQuadrant(damageSource, c, diameter, damageAmount * 0.25f, Quadrant.SW);
    }
}
```
```cs
void DamageQuadrant(GameObject damageSource, float damageAmount, 
                    Point c, int diameter, Quadrant quadrant)
{
    foreach (ShipModule m in EnumQuadrant(c, diameter, quadrant))
        if (m.DamageExplosive(damageSource, ref damageAmount))
            return true;
}
```
The extra code for distance falloff can be simplified and also the iteration logic can then be simplified. When I look at the current `EnumModulesQuadrants` it's really difficult to understand compared to the previous simplified approach, so breaking this code into smaller pieces could help.

The proposed `EnumQuadrant` function would only deal with one purpose: the logic on how to iterate the modules inside its quadrant. The whole algorithm should be able to be generalized via `Point direction = GetDirectionFromQuadrant(quadrant);` -- with NW being {X:-1,Y:-1}, NE being {X:+1,Y:-1} etc.

I'm not really sold on having separate Diagonal damage transfer either. The quadrant logic should simply keep expanding by 1 distance, first doing horizontal, then vertical tiles, and finally the single diagonal piece. We always know the number of potential tiles affected, which will grow like this:
```
3 3 3 3
3 2 2 2
3 2 1 1
3 2 1 x
```
For iteration 0, there is always 1 module.
For iteration 1, there are always 3 modules.
For iteration 2, there are always 5 modules.
For iteration 3, there are always 7 modules.

So we can divide the damage by 1, 3, 5, 7, 9, ...; And this will naturally spread out the damage in a somewhat linear fashion without using the old radial falloff logic that made balancing impossible.

Ship_Game/Ships/Ship_ModuleGrid.cs
gkapulis: actually it should be spread per quadrant here as well, if the root module died and left damage for the explosion. I'll fix this.

PR #143, opened 2024-03-22
Feature/fleet generation improvements

SmischenkoB:


Ship_Game/GameScreens/FleetDesign/FleetDesignScreen_Draw.cs
```diff
@@ -355,8 +356,7 @@ void DrawSelectedData(SpriteBatch batch, DrawTimes elapsed)
                 RequisitionForces.Visible = true;
                 SaveDesign.Visible = true;
                 LoadDesign.Visible = true;
-                if (f.Ships.Count > 0 )
-                    AutoArrange.Visible = true;
+                AutoArrange.Visible = f.DataNodes.Where(n => n.Ship == null && n.Goal == null).Count() == 0;
```
gkapulis, line 359: Seems This is not related to this PR

game/Content/FleetBuildRatiosRandomization.yaml
```diff
@@ -0,0 +1,83 @@
+# This will set the Fleet Randomized Ratios for the AI ship build logic.
```
gkapulis, line 1: This file is not needed.
In the fleetRatios.yaml, you can use Range struct instead of duplicating a yaml file.
check Range.cs. and its implementation in PlanetTypes.yaml (PopPerTile, for instance)
The Range struct has the Generate method which will generate a random number for the provided range. 
It can also get only the minimum or the maximum numbers, incase a modder does not want random numbers.

gkapulis (changes requested):
That would be an interesting change indeed, but we need a better implementation of this.
Check my comments.
Also, please add flag in [Globals.yaml ](https://github.com/TeamStarDrive/StarDrive/blob/main/game/Content/Globals.yaml)which if true, uses the random numbers and if not, gets the maximum (or minimum) number in the range struct (choose what you want, just add a text to explain it in the flag).
For instance - check `DisableRemnantStory` flag in Globals.yaml or any of the other feature flags there, Remember to add the new flag to the BinarySerializationTests.cs as well.

Ship_Game/GameScreens/FleetDesign/FleetDesignScreen_Draw.cs
gkapulis, line 359:
```suggestion
     AutoArrange.Visible = !f.DataNodes.Any(n => n.Ship == null && n.Goal == null);
```

Try avoid using LINQ. it is slow as F*** ,  we have faster implementations for these.   its not so severe in the UI, but lets keep it ordered. :)

SmischenkoB: yes it works if all ships are built, so no templates just complete ships

game/Content/FleetBuildRatiosRandomization.yaml
SmischenkoB: done

game/Content/Globals.yaml
```diff
@@ -52,6 +52,7 @@ Globals:
   DisableRemnantStory: false
   DisablePirates: false
   AIUsesPlayerDesigns: true
+  EnableRandomizedAIFleetSizes: true
```
gkapulis, line 55: Add here  a comment explaining what it does. Also add it to example mod global.yaml , maybe.

Ship_Game/Empire.cs
```diff
@@ -2350,6 +2350,11 @@ public bool KillsForRemnantStory(Empire they, Ship killedShip)
             return true;
         }
 
+        public bool IsRandomizedAiFleetSizesEnabled()
+        {
+            return Universe.P.EnableRandomizedAIFleetSizes;
+        }
+
```
gkapulis, line 2357: remove this. and use that flag directly from SetFleetRatios();
like `SetFleetRatios(empire.universe.P....);`

gkapulis (commented):
Almost done. 
Did you check EmpireAI tests pass? There is a test there which checks the fleet ratio i think and if its random, it should fail. You will need to set the flag to false before running this test and maybe update the expected results for your new minimum values.

Ship_Game/AI/StrategyAI/FleetRatios.cs
```diff
@@ -48,23 +48,23 @@ public FleetRatios(Empire empire)
...
             MinSupport     = 0;
             MinCombatFleet = 0;
-            SetFleetRatios();
+            SetFleetRatios(empire.Universe.P.UseRandomRatios);
 
 
         }
 
-        public void SetFleetRatios()
+        public void SetFleetRatios(bool IsRandomizedAIFleetSizes)
```
gkapulis, line 56:
```suggestion
        public void SetFleetRatios(bool useRandomFleetSizes)
```

Ship_Game/Data/ResourceManager.cs
```diff
@@ -2004,11 +2005,22 @@ static void LoadSunZoneData()
...
-            return BuildRatios[canBuild];
+            int[] fleetSize = new int[BuildRatios[canBuild].Length];
+            SeededRandom seededRandom = new SeededRandom();
+
+            for (int i = 0; i < BuildRatios[canBuild].Length; i++)
+            {
+                fleetSize[i] = IsRandomizedAIFleetSizes 
+                    ? (int) Math.Floor(BuildRatios[canBuild][i].Generate(seededRandom))
+                    : (int) BuildRatios[canBuild][i].Min;
+            }
```
gkapulis, line 2021: actually you dont need this loop and an if condition for each member.
return a Range[] and deal with the flag in `SetCounts`

gkapulis, line 2021:
```suggestion
        public static Range[] GetFleetRatios(BuildRatio canBuild)
        {
            return BuildRatios[canBuild];
        }
            }

```

Ship_Game/Universe/UniverseParams.cs
```diff
@@ -75,6 +75,7 @@ public class UniverseParams
     [StarData] public bool FilterOldModules;
 
     [StarData] public bool DisableRemnantStory;
+    [StarData] public bool UseRandomRatios;
```
gkapulis, line 78:
```suggestion
    [StarData] public bool EnableRandomizedAIFleetSizes
```

Ship_Game/Universe/UniverseParams.cs
```diff
@@ -100,6 +101,7 @@ public UniverseParams()
         StartingPlanetRichnessBonus = s.StartingPlanetRichnessBonus;
         GravityWellRange = s.GravityWellRange;
         DisableRemnantStory = s.DisableRemnantStory;
+        UseRandomRatios = s.EnableRandomizedAIFleetSizes;
```
gkapulis, line 104:
```suggestion
        EnableRandomizedAIFleetSizes = s.EnableRandomizedAIFleetSizes;
```

Ship_Game/AI/StrategyAI/FleetRatios.cs
```diff
@@ -48,23 +48,23 @@ public FleetRatios(Empire empire)
...
-            else if (OwnerEmpire.CanBuildCorvettes)   counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildCorvettes);
-            else                                      counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildFighters);
+            if      (OwnerEmpire.CanBuildCapitals)    counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildCapitals, IsRandomizedAIFleetSizes);
+            else if (OwnerEmpire.CanBuildBattleships) counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildBattleships, IsRandomizedAIFleetSizes);
+            else if (OwnerEmpire.CanBuildCruisers)    counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildCruisers, IsRandomizedAIFleetSizes);
+            else if (OwnerEmpire.CanBuildFrigates)    counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildFrigates, IsRandomizedAIFleetSizes);
+            else if (OwnerEmpire.CanBuildCorvettes)   counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildCorvettes, IsRandomizedAIFleetSizes);
+            else                                      counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildFighters, IsRandomizedAIFleetSizes);
```
gkapulis, line 67:
```suggestion
        public void SetFleetRatios(bool useRandomRatios)
        {
            // fighters, corvettes, frigate, cruisers, capitals, troopShip,bombers,carriers,support

            Range[] counts;

            if      (OwnerEmpire.CanBuildCapitals)    counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildCapitals);
            else if (OwnerEmpire.CanBuildBattleships) counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildBattleships);
            else if (OwnerEmpire.CanBuildCruisers)    counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildCruisers);
            else if (OwnerEmpire.CanBuildFrigates)    counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildFrigates);
            else if (OwnerEmpire.CanBuildCorvettes)   counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildCorvettes);
            else                                      counts = ResourceManager.GetFleetRatios(BuildRatio.CanBuildFighters);

            SetCounts(counts, useRandomRatios);
            CountIndexed = counts;
        }
```

So you just return the range from the static function and there are no loops involved. 
In SetCounts you can use either min or Generate depends on the flag

gkapulis (approved):

PR #163, opened 2024-07-14
Update GameText.yaml

irk72ag:
Added missing Russian localization. The noted errors/typos have been proofread and corrected. The work will continue until all lines are fully translated


gkapulis (commented):
Thanks for the contribution!
The target branch should mainly be mars-1.50 and not main.

PR #203, opened 2024-10-05
Feature/dyson swarm

gkapulis:
- Added Dyson Swarm build options and Techs.
- Added 2 more Binary Suns
- Added 1 Purple Star
- Added 1 Barren Brown Dwarf


Ship_Game/GameScreens/ColonyScreen/ColonyScreen_Update.cs
```diff
@@ -82,6 +84,55 @@ void UpdateTradeTab()
...
+            {
+                if (DysonSwarmTypeTitle?.Visible == true)
+                    HideDysonSwarmUI();
+
+                return;
+            }
+            DysonSwarmTypeTitle.Visible       =
+            DysonSwarmControllerPanel.Visible =
+            DysonSwarmProdBoost.Visible       =
+            DysonSwarmPanel.Visible           = true;
```
SmischenkoB, line 99: I understand that everything should be true here, but I think true on each row would be better

Ship_Game/GameScreens/ColonyScreen/ColonyScreen_Update.cs
```diff
@@ -82,6 +84,55 @@ void UpdateTradeTab()
...
+        void HideDysonSwarmUI()
+        {
+            DysonSwarmTypeTitle.Visible =
+            DysonSwarmControllerPanel.Visible =
+            DysonSwarmProdBoost.Visible =
+            DysonSwarmPanel.Visible = 
+            DysonSwarmStartButton.Visible =
+            DysonSwarmStatus.Visible = 
+            DysonSwarmKillButton.Visible = false;
+            DysonSwarmOverclock.Visible = false;
```
SmischenkoB, line 133: Can I ask to put false on each row here

Ship_Game/Universe/SolarBodies/DysonSwarm.cs
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        public void DeploySwarmSat()
+        {
+            NumSwarmSats = (NumSwarmSats + 1).UpperBound(RequiredSwarmSats);
+        }
+      
+        void LoadDysonRings()
+        {
+            int neededRings = (int)(SwarmCompletion * 100) / 10;
```
SmischenkoB, line 145: wouldn't this line turn int back to float? Maybe use some round function? Or do casting to int in the end?

Ship_Game/Universe/UniverseScreen/UniverseScreen.cs
```diff
@@ -206,20 +206,22 @@ void ResetLighting(bool forceReset)
             AddLight("Global Back Light", new Vector2(0, 0), 0.6f, globalLightRad, Color.White, +globalLightZPos, fillLight: false, shadowQuality: 0f);
 
             foreach (SolarSystem system in UState.Systems)
```
SmischenkoB, line 208: there seems nothing inside this foreach

game/Content/GameText.yaml
```diff
@@ -3554,10 +3554,10 @@ BB_Tech_TachyonScanning_SensorRangeBonus_Bonus:
...
 BB_Tech_Terraforming_Name:
  Id: 683
- ENG: "Terraforming"
- RUS: "Терраформинг"
- SPA: "Terraformación"
- UKR: "Тераформування"
+ ENG: "Terraform"
+ RUS: "Терраформировать"
+ SPA: "Terraformar"
+ UKR: "Тераформа"
```
SmischenkoB, line 3560: if it's verb, than UKR: "Тераформувати",  if noun UKR: "Тераформер"

Ship_Game/Universe/SolarBodies/DysonSwarm.cs
gkapulis: SwarmCompletion is float. It gets converted to 1-100 and to Int. Int divided by int gives a floored int.
so if the result is 95/10 = 9.5 which is int 9. This is what i want here.

Ship_Game/Universe/UniverseScreen/UniverseScreen.cs
gkapulis: It was not changed so you see  trailing code after the actual change:
this is the code:
```
           foreach (SolarSystem system in UState.Systems)
                ResetSolarSystemLights(system);
```

PR #212, opened 2025-01-27
Influence Fix and refactor

gkapulis:
- Remove border planets from influence, Instead, combine colony influence to the system itself, saving some CPU time and fixing under the hood dynamic influence of planets orbiting sun while the UI displayed fixed influence to the player.
- Colony projection is now determined by empire projection and not related to pop anymore. 
- Removed a feature flag related to this: `UsePlanetaryProjection`
- Removed projection range xml tags from building as they no longer set the projection.

### DIFF
--- a/Ship_Game/Empire_Borders.cs
+++ b/Ship_Game/Empire_Borders.cs
@@ -7,6 +7,7 @@ using Ship_Game.Gameplay;
 using Ship_Game.Universe;
 using Ship_Game.AI;
 using Ship_Game.Data.Serialization;
+using System.Linq;
 
 namespace Ship_Game;
 
@@ -44,7 +45,6 @@ public sealed partial class Empire
     public EmpireFirstContact FirstContact = new();
 
     readonly Array<InfluenceNode> OurBorderSystems = new(); // all of our systems
-    readonly Array<InfluenceNode> OurBorderPlanets = new(); // all of our planets
     readonly Array<InfluenceNode> OurBorderShips = new(); // SSP-s and some Bases
 
     readonly Array<InfluenceNode> OurSensorPlanets = new(); // all of our planets
@@ -62,7 +62,6 @@ public sealed partial class Empire
         OurSensorShips.Clear();
         OurSensorPlanets.Clear();
         OurBorderShips.Clear();
-        OurBorderPlanets.Clear();
 
         BorderNodes = Empty<InfluenceNode>.Array;
         SensorNodes = Empty<InfluenceNode>.Array;
@@ -212,7 +211,11 @@ public sealed partial class Empire
         bool known = empireKnown || planet.IsExploredBy(Universe.Player);
 
         // NOTE: planets always provide influence
-        OurBorderPlanets.Add(new(planet, planet.GetProjectorRange(), known));
+
+
+        if (!OurBorderSystems.Any(n => n.Source == planet.System))
+            OurBorderSystems.Add(new(planet.System, GetProjectorRadius(), known));
+
         OurSensorPlanets.Add(new(planet, planet.SensorRange, empireKnown));
     }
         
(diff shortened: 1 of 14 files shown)


Ship_Game/Empire_Borders.cs
```diff
@@ -274,9 +277,9 @@ void UpdateSensorAndBorderRadiuses()
         {
             n.Radius = useSensorRange ? ((Ship)n.Source).SensorRange : projectorRadius;
         }
-        foreach (ref InfluenceNode n in borderPlanets)
+        foreach (ref InfluenceNode n in borderSystems)
         {
-            n.Radius = ((Planet)n.Source).GetProjectorRange();
+            n.Radius = GetProjectorRadius();
```
RedFox20, line 282: Could have reused `float projectorRadius` from above to improve perf

Ship_Game/Universe/InfluenceTree.cs
```diff
@@ -54,36 +54,30 @@ public struct InfluenceObj
...
                 float radius = projector.Loyalty.GetProjectorRadius();
                 return (source.Position, radius, radius);
             }
 
-            if (source is Planet planet)
+            if (source is SolarSystem system)
             {
-                if (planet.System.Position == Vector2.Zero)
+                if (system.Position == Vector2.Zero)
                     Log.Error("InfluenceTree: Planet.ParentSystem position is Zero!");
```
RedFox20, line 68: System position is Zero

Ship_Game/Universe/SolarBodies/Planet/Planet_Generate.cs
```diff
@@ -185,7 +185,7 @@ public void GenerateNewHomeWorld(RandomBase random, Empire owner, SolarSystemDat
             InitPlanetType(type, scale, fromSave: false);
             SetOwner(owner);
             IsHomeworld = true;
-            Owner.SetCapital(this);
+            owner.SetCapital(this);
```
RedFox20, line 188: I think there was a subtle reason for this `Owner` versus `owner`

Ship_Game/Universe/UniverseState.cs
```diff
@@ -615,13 +615,13 @@ public void OnPlanetOwnerAdded(Empire owner, Planet planet)
             if (planet.Budget?.Owner != owner)
                 planet.CreatePlanetBudget(owner);
             owner!.AddBorderNode(planet);
-            Influence.Insert(owner, planet);
+            Influence.Insert(owner, planet.System);
```
RedFox20, line 618: This introduces a massive bug, previously a planet was carefully checked so it would only be added once.
And so, there is no protection to adding multiple influences for the same entity in Influence tree Insert. You can check the Insert() function and can see that it simply inserts the object without any protective checks.

So you should add protection inside the Insert() - which shouldn't be an issue for performance tbh. You can use debug visualizer to see if multiple influences are added.

Ship_Game/Empire_Borders.cs
```diff
@@ -211,8 +209,10 @@ public void AddBorderNode(Planet planet)
         bool empireKnown = IsThisEmpireKnownByPlayer();
         bool known = empireKnown || planet.IsExploredBy(Universe.Player);
 
-        // NOTE: planets always provide influence
-        OurBorderPlanets.Add(new(planet, planet.GetProjectorRange(), known));
+        // NOTE: planets always provide influence and actually project it from system center.
+        if (!OurBorderSystems.Any(n => n.Source == planet.System))
+            OurBorderSystems.Add(new(planet.System, GetProjectorRadius(), known));
```
SmischenkoB, line 214: if two empire have planets in one system will it affect eachother?

game/Content/GameText.yaml
```diff
@@ -15108,7 +15108,7 @@ BluePrintsOverView:
  ENG: "Colony Blueprints are mainly lists of buildings which can be uploaded for the Governor to follow. The Governor will prioritize these building if possible, and in case the Blueprints are Exclusive, the Governor will build only the specified buildings in the uploaded Blueprints. Blueprints can also be linked to other Blueprints so when they finished, new Blueprints are uploaded."                   
 BluePrintsEnableGovernorToLoad:
  Id: 6218
- ENG: "Enable a Governor in order upload Blueprints."                    
+ ENG: "Enable a Governor in order to upload Blueprints."
```
SmischenkoB, line 15111: please, check if you have latest master to resolve potential conflict with my localization PR

PR #299, opened 2026-06-14
Localization: add missing Russian (RUS) translations

wCELENw:
## Summary
Adds Russian (`RUS`) translations for **798** entries in `game/Content/GameText.yaml` that previously had no `RUS` line. Russian coverage in this file goes from 2794 → **3592 (100% of entries)**.

- **597** entries reuse community Russian translations shipped with *StarDrivePlus 1.51*.
- **201** entries are newly translated (no prior `RUS`), with terminology aligned to the file's existing Russian strings (via an extracted glossary of ~1.6k accepted terms).

## Scope & safety
- **Purely additive: `+798 / -0`**, single file.
- Only `RUS` lines were inserted (immediately after each `ENG`). No keys, `Id`s, `ENG`, or other languages (`SPA` / `UKR` / `PTB` / `GER`) were touched.
- Inserted values preserve formatting tokens, placeholders, line breaks and meaningful spacing from the source.

Happy to adjust terminology or re-run through the in-repo localizer (`--run-localizer`) if preferred. A separate proofreading PR for *existing* Russian strings can follow.

### DIFF
--- a/game/Content/GameText.yaml
+++ b/game/Content/GameText.yaml
@@ -470,6 +470,7 @@ Mercantile:
 MercantileRacesAreNaturalTraders:
  Id: 67
  ENG: "Mercantile races are natural traders, and gain 0.5 Credits for each unit of food or production moved via freighter within the empire, they also have improved supply chains for trade agreements which can help supply the demand of larger empires having trade agreements with them."
+ RUS: "Меркантильные расы — прирожденные торговцы: они получают 0,5 кредита за каждую единицу еды или продукции, перевезённую грузовым судном внутри империи; также у них улучшены цепочки поставок по торговым договорам, что помогает удовлетворять спрос крупных империй, имеющих с ними соглашения."
  UKR: "Меркантильні раси є природними торговцями та отримують 0,5 кредиту за кожну одиницю їжі чи продукції, що переміщується через вантажне судно в межах імперії; вони також мають покращені ланцюжки постачання для торгових угод, які можуть допомогти задовольнити попит великих імперій, які мають з ними торгові угоди."
  PTB: "Raças mercantis são comerciantes naturais e recebem 0,5 Crédito para cada unidade de comida ou produção transportada por cargueiro dentro do império. Elas também possuem cadeias de suprimento melhores para acordos comerciais, o que pode ajudar a suprir a demanda de impérios maiores que tenham acordos comerciais com elas."
 Wasteful:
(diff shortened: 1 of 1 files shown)


gkapulis:
Thank you for this, and sorry it sat so long. The branch had drifted too far to merge cleanly, so rather than lose the work I imported the content in-house: **[29ae80e19](https://github.com/TeamStarDrive/StarDrive/commit/29ae80e192d3f1d157032659432c216fd35975f6)** on `jupiter-1.60/fixes_33`, with you as `Co-Authored-By`.

All 798 entries you translated are in. Russian coverage in `GameText.yaml` goes from 2695 to 3493 of 3722.

A few notes on what changed on the way in, in case it is useful for future passes:

- **Em dash and non-breaking hyphen.** Every `*.spritefont` in the repo declares only `U+0020-007E`, `U+00A0-00FF` and `U+0400-045F`, with `?` as the default character. `—` (U+2014) and `‑` (U+2011) are outside that, so they would have drawn as a literal `?` in game. Replaced with a plain `-` in 45 values. Worth knowing for any future translation work here — guillemets `«»` and `ё` are fine, they fall inside the Latin-1 and Cyrillic regions.
- **Espionage notification line breaks.** Twelve of the `InfiltrationDisrupt*` strings dropped the `\n` from the English. `NotificationManager.Draw` uses `batch.DrawString` with no wrapping and right-aligns from the icon (`cursor.X = ClickRect.X - msgSize.X - 3f`), so a long unbroken line runs off the left edge of the screen. I re-broke those twelve at clause boundaries.
- **Eight tooltips were re-translated**, not because of the Russian but because their English has been rewritten since June as part of an in-game Codex content pass — the refining-time tooltip now mentions Production for cybernetic races, the rebellion op targets the *least* populated colony, the biosphere tile scraps on a plain click rather than a right click.
- **Three Dyson entries** said «Сфера Дайсона»; the game's tech is a Swarm, and your other nine Dyson strings correctly say «Рой Дайсона», so I made those three match.

I also deliberately did **not** restore Russian on 98 entries that had it before: their English was rewritten in the same Codex pass, so the old translations now describe rules that no longer exist.

229 entries still have no Russian, nearly all of them Codex help text added after you opened this. If you would like another pass at those, it would be very welcome — and please do open it against the current `jupiter-1.60/fixes_NN` branch so it does not go stale again.

Closing this as landed. Thanks again.

gkapulis:
Landed in-house as 29ae80e19 with credit; see the previous comment for details.

PR #348, opened 2026-07-22
Fix fertility displayed as -0 (#325)

### LINKED ISSUE #325: On some planets fertility is shown as -0
Image

Image

Fertility cannot go below 0. If a planet with 0 fertility builds a negative fertility building, then fertility remains 0 and it just means a higher hurdle for positive fertility buildings to get it above 0 again. So, there is some virtual counting below 0. But there seems no reason to have a 0 and -0 fertility. It is just confusing.

Ludoal:
`UpdateBaseFertility`'s decrease branch could write a small negative `BaseFertility`: `Clamped(0, BaseFertility - 0.01f)` returns the max bound when `BaseFertility` is between 0 and 0.01, so the value went below the documented `>= 0` floor and the UI rendered it as "-0" (both reported screenshots are that transient, caught during the one planet-turn it lasts).

The fix decreases by the same 0.01 step, floored at `totalFertility` but never below 0. The negative building deficit (the "virtual counting below zero" described in the issue) still lives in `totalFertility`, so the gameplay hurdle is unchanged.

Fixes #325

### DIFF
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet.cs
@@ -1108,7 +1108,7 @@ namespace Ship_Game
             if (BaseFertility < totalFertility)
                 BaseFertility = (BaseFertility + 0.01f).Clamped(0, totalFertility); // FB - Slowly increase fertility to max fertility
             else if (BaseFertility > totalFertility)
-                BaseFertility = BaseFertility.Clamped(0, BaseFertility - 0.01f); // FB - Slowly decrease fertility to max fertility
+                BaseFertility = (BaseFertility - 0.01f).LowerBound(totalFertility.LowerBound(0)); // FB - Slowly decrease fertility to max fertility, never below 0
         }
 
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified against the source:

- Confirmed the root cause: `Clamped` returns the max bound when `value >= max - epsilon`, so `BaseFertility.Clamped(0, BaseFertility - 0.01f)` yields a small negative for one planet-turn when `0 < BaseFertility < 0.01` — exactly the transient -0 in both issue screenshots.
- Confirmed the fix preserves the negative-buildings hurdle: the deficit is tracked in `BuildingsFertility` (signed, serialized, never clamped), and `BaseFertility` merely chases `BaseMaxFertility + BuildingsFertility`. Positive buildings pay down the deficit through the sum; they cannot bypass it. Traced build/scrap symmetry in Planet_ConstructionQueue as well.
- Bonus: the old decrease branch could undershoot the target and oscillate one turn via the increase branch; the new `LowerBound(totalFertility)` floor converges faster. Also self-heals saves that stored the transient negative.

Nice catch, thanks!

PR #349, opened 2026-07-22
Sort fleet button ship icons big-to-small (#318)

### LINKED ISSUE #318: The order in which ships are shown on the main screen is inconsistent, making it harder to compare your fleets
Image

The order in which ships are shown on the main screen is inconsistent, making it harder to compare your fleets. Sometimes the smaller ships are shown first, sometimes the larger ships are shown first. It would probably be best to show the ships from big to small, that is titans, battleship, cruisers, frigates, corvettes, fighters and drones.

In the screenshot, you can see that in fleet 1 and 3, the corvettes are depicted first and the frigates second, while in fleet 3, the frigates are depicted first and the corvettes second.

Ludoal:
Both fleet button icon modes (per-ship icons and grouped sums) iterated `fleet.Ships` in raw list order, so the icon order differed between fleets and made them hard to compare at a glance.

Ships are now sorted by `DesignRole` descending (capital … fighter, matching the enum order) with design name as tie-breaker, so every fleet shows its ships big-to-small in a stable order — the ordering suggested in the issue.

Fixes #318

### DIFF
--- a/Ship_Game/GameScreens/FleetDesign/FleetButton.cs
+++ b/Ship_Game/GameScreens/FleetDesign/FleetButton.cs
@@ -121,10 +121,12 @@ public class FleetButton : UIPanel
     void DrawFleetShipIcons30(SpriteBatch batch, Fleet fleet, float x, float y)
     {
         // Draw ship icons to right of button
+        // sort by role, big ships first, so the order is consistent between fleets
+        Ship[] ships = fleet.Ships.Sorted(s => (-(int)s.DesignRole, s.Name));
         Vector2 shipSpacingH = new(x, y);
-        for (int i = 0; i < fleet.Ships.Count; ++i)
+        for (int i = 0; i < ships.Length; ++i)
         {
-            Ship ship = fleet.Ships[i];
+            Ship ship = ships[i];
             RectF iconHousing = new(shipSpacingH.X, shipSpacingH.Y, 15, 15);
             shipSpacingH.X += 18f;
             if (shipSpacingH.X >= x + 180) // 10 Ships per row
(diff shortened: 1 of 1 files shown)


gkapulis (changes requested):
Thanks — the cross-fleet inconsistency this targets is real, and a stable sort is the right shape. Two changes needed before we take it.

**1. `DesignRole` descending isn't a size ordering.**

`RoleName` (Ship_Game/Ships/RoleName.cs) isn't ordered by hull size end-to-end. Descending `(int)DesignRole` gives:

`prototype(24), capital(23), battleship, cruiser, destroyer, frigate, corvette, drone(17), gunboat, scout, fighter(14), carrier(13), bomber(12), support(11), troopShip(10), troop, freighter, ...`

So the pure warship line capital→corvette does match issue #318 exactly (including the corvette-vs-frigate case in the screenshot), but:

- **`carrier`, `bomber`, `support` and `troopShip` all sort BELOW `fighter`** — a carrier is typically one of the largest hulls in a fleet, and it would draw after the fighters and drones. That is the same "why is the big ship at the end" complaint the issue is about.
- `drone`/`gunboat`/`scout` outrank `fighter`, inverting the issue's requested "fighters and drones" tail.
- `prototype` outranks `capital`.

Please sort on an actual size proxy instead — `s.SurfaceArea` descending with the design name as tie-breaker. That is literally the "big to small" the issue asks for, it puts carriers and troop transports where players expect them, and it still keeps same-role ships adjacent in practice (same hull → same tactical icon → similar surface area).

**2. In `DrawFleetShipIconsSums` the sort has no defined effect, and it is the expensive one.**

There, the sorted ships are only used to populate `Map<TacticalIcon, int> sums`, and the drawing loop iterates `sums.Keys.ToArr()`. `Map<K,V>` derives from `Dictionary<K,V>` (SDUtils/Map.cs:22), whose key enumeration order is not contractually insertion order — it happens to hold here because nothing is removed, but the display order is incidental rather than something the sort guarantees.

It is also the costly path: this branch only runs for fleets with more than 30 ships, so it sorts 30–100+ ships every frame in order to arrange at most 12 icon groups.

Please build the sums from the unsorted list as before, and sort the icon groups themselves (≤12 entries) right before the draw loop. That makes the ordering explicit instead of relying on dictionary internals, and drops the per-frame work by an order of magnitude on big fleets.

**Minor, related:** `Sorted()` allocates two arrays per call (the `ToArr` copy plus the `keys` array in CollectionExt.Sort), once per fleet button per frame in the draw path. Fixing (2) removes the worst of it; if the ≤30 path is easy to make allocation-free too, even better, but we won't hold the PR on that.

The threading exposure is unchanged (reading sim-mutated `fleet.Ships` from the draw thread), so no concern there — same as the existing indexed loop.

PR #350, opened 2026-07-22
Resolve fleet requisition goals while the Fleet Manager is open (#320)

### LINKED ISSUE #320: Ship requisitions in the fleet manager show 'Need spaceport'
Image

Ship requisitions in the fleet manager show 'Need spaceport' when you first requisition ships. When you leave the fleet manager and return, the locations where the ships will be constructed are shown. It seems similar to the issue when you delete ships in the ship array and it doesn't show a status update that it is scrapping in planet xyz.

Ludoal:
The Fleet Manager is a fullscreen (non-popup) screen: it hides the `UniverseScreen`, whose `Draw()` is what signals `DrawCompletedEvt` to the simulation thread. With the universe hidden the sim thread parks, so the requisition goals queued via `AddGoalAndEvaluate` never ran and every node showed "Need spaceport" until the screen was closed — which also explains the "leave and return and it looks right" behavior in the report.

This adds `UniverseScreen.PumpPendingSimThreadActions()`: when the universe is hidden and the game paused, it signals the sim loop so it runs its paused fast-path (pending actions only, simulation time never advances), and `FleetDesignScreen.Update` calls it. Requisitioned nodes now switch to "Building at: <planet>" while you watch.

Fixes #320

### DIFF
--- a/Ship_Game/GameScreens/FleetDesign/FleetDesignScreen_Update.cs
+++ b/Ship_Game/GameScreens/FleetDesign/FleetDesignScreen_Update.cs
@@ -10,6 +10,11 @@ namespace Ship_Game
     {
         public override void Update(float fixedDeltaTime)
         {
+            // This screen hides the universe, which parks the sim thread and would leave
+            // requisition goals frozen in the pending queue (they showed "Need spaceport"
+            // until the screen was closed). Pump the queue so they resolve while we watch.
+            Universe.PumpPendingSimThreadActions();
+
             CamPos.X = CamPos.X.SmoothStep(DesiredCamPos.X, 0.2f);
             CamPos.Y = CamPos.Y.SmoothStep(DesiredCamPos.Y, 0.2f);
             CamPos.Z = CamPos.Z.SmoothStep(DesiredCamPos.Z, 0.2f);
--- a/Ship_Game/Universe/UniverseScreen/UniverseScreen.UpdateGame.cs
+++ b/Ship_Game/Universe/UniverseScreen/UniverseScreen.UpdateGame.cs
@@ -394,6 +394,19 @@ namespace Ship_Game
         }
 
         /// <summary>
+        /// Wakes up the parked simulation thread so queued RunOnSimThread actions still
+        /// execute while a fullscreen game screen hides the universe (normally Draw()
+        /// signals the sim loop, but a hidden universe never draws). The UState.Paused
+        /// guard means only the paused fast-path runs: pending actions are invoked,
+        /// simulation time never advances.
+        /// </summary>
+        public void PumpPendingSimThreadActions()
+        {
+            if (UState.Paused && !Visible)
+                DrawCompletedEvt.Set();
+        }
+
+        /// <summary>
         /// Queues action to run on the Simulation thread, aka ProcessTurns thread.
         /// </summary>
         public void RunOnSimThread(Action action)


gkapulis (changes requested):
The diagnosis is exactly right and I verified the whole chain: `GameScreen.PreUpdate` sets `Visible = ScreenState != ScreenState.Hidden` (GameScreen.cs:366), a non-popup screen sets `coveredByOtherScreen` for everything below it (ScreenManager.cs:796), so the Fleet Manager drives `UniverseScreen.Visible` false, `ScreenManager.Draw` skips it, `DrawCompletedEvt` is never signalled, and the sim thread parks on `WaitOne` with the marshaled requisition goals stuck in the queue. `UState.Paused` is genuinely satisfied too, since `FleetDesignScreen` is constructed with `base(u, toPause: u)` and the `GameScreen` ctor sets it (GameScreen.cs:134-138). Good find.

The fix wants one change though: **signal when an action is enqueued, not every frame.**

**Why the per-frame pump buys nothing.** The paused fast-path doesn't re-evaluate goals — it only runs `InvokePendingSimThreadActions()`, `Objects.Update(Zero)` and `UpdateClickableItems()`. Goal evaluation happens inside the queued action itself: `EmpireAI.AddGoalAndEvaluate` marshals `() => { GoalsList.Add(goal); goal.Evaluate(); }`, so the goal is evaluated exactly once, when the queue is drained. Pumping at frame rate gives an identical result to pumping once per enqueue; the other ~59 wakes per second just redo list bookkeeping on a universe that cannot have changed.

**Why `RunOnSimThread` is the right place.** It is the single universal enqueue point, and it already enqueues before it would signal, so the ordering is correct by construction:

```csharp
public void RunOnSimThread(Action action)
{
    if (action != null)
    {
        PendingSimThreadActions.Enqueue(action);
        if (!Visible && UState.Paused)
            DrawCompletedEvt.Set(); // a hidden universe never draws, so wake the sim to drain this
    }
    else
    {
        Log.WarningWithCallStack("Null Action passed to RunOnEmpireThread method");
    }
}
```

That drops the `FleetDesignScreen_Update.cs` change entirely, and it fixes the whole class rather than one screen — Research, Ship Design, Planet List and Empire Management all starve the sim the same way, and any marshaled action they queue is equally frozen until the screen closes.

No lost-wakeup risk: `PendingSimThreadActions` is a `SafeQueue`, `InvokePendingSimThreadActions` drains in a `while` loop, and if more actions arrive mid-drain their own `Set` leaves the `AutoResetEvent` signalled for the next pass.

**Please keep the guard as-is.** Always signalling on enqueue looks tempting but is not free: an extra pass through the *unpaused* branch runs `AutoAdjustSimulationFrameRate()` and the `AutoRampGameSpeed` block, which does `++UState.GameSpeed` / `--UState.GameSpeed`. Signalling on every enqueue during normal play would perturb the auto-speed ramp. `!Visible && UState.Paused` confines the wake to exactly the parked case.

Keep the `<summary>` doc — it documents a genuinely non-obvious threading contract — just reword it for the enqueue-driven version. The public `PumpPendingSimThreadActions()` method can go away with the call site.

One thing we'll want to play-test after this lands, not a change request: the Fleet Manager previously ran with the sim thread parked, i.e. exclusive access to fleet data. Waking it restores the same UI/sim concurrency the paused universe screen has always had, but the Fleet Manager mutates fleet structure (node add/remove, ship assignment, requisition) more aggressively than other paused-state UI, so it is worth a drag-heavy session to confirm nothing tears.

PR #351, opened 2026-07-22
Refresh Ships Array status after a scrap order actually runs (#322)

### LINKED ISSUE #322: When you delete a ship in the ship array, it doesn't update its status
This is a more recently introduced bug. When you used to delete a ship in the ship array, then it would get a new status that it is scrapping ship in planet xyz. Now, there is no status update as if you never gave a command to scrap. When you exit the ship array and re-enter it, then it shows the location where the ship is being scrapped. It seems similar to the issue when you requisition ships in the fleet manager and it shows 'need spaceport' instead of where the ship is being constructed.

Ludoal:
`OrderScrapShip` defers its `ScrapShip` goal to the sim thread via `AddGoalAndEvaluate`, so the immediate `GetStatusText` refresh in the click handler still saw the pre-scrap AI state and the row kept its old status until the screen was reopened.

Queue a `ResetStatus` on the sim thread right after the order — the same pattern the shift-click `MassScrap` branch already uses. (The reporter's hunch that this is the same family as the fleet-requisition "Need spaceport" issue was right: both are UI reads racing the deferred sim-thread goal, see #320.)

Fixes #322

### DIFF
--- a/Ship_Game/ShipListScreenItem.cs
+++ b/Ship_Game/ShipListScreenItem.cs
@@ -405,7 +405,10 @@ namespace Ship_Game
                         }
                         else
                         {
+                            // OrderScrapShip defers the ScrapShip goal to the sim thread,
+                            // so refresh the status only after the goal has actually run
                             Ship.AI.OrderScrapShip();
+                            Screen.Universe.RunOnSimThread(() => Screen.ResetStatus());
                         }
                     }
                     StatusText = GetStatusText(Ship);


gkapulis (changes requested):
Diagnosis is exact and the fix does what it says. One latent race I'd like closed while we're in here, then this is good to merge.

## Verified

The deferral chain is real: `OrderScrapShip()` → `AddScrapShipGoal` → `AddGoalAndEvaluate`, which marshals `GoalsList.Add(goal); goal.Evaluate();` to the sim thread (`EmpireAI.cs:563-564`). `AI.State` does not become `Scrap` until that action runs, so the synchronous `StatusText = GetStatusText(Ship)` at `ShipListScreenItem.cs:411` reads pre-scrap state. Exactly as described.

The fix is not invented either — it mirrors the shift-click `MassScrap` branch four lines above (`:403-404`) and relies on the same FIFO ordering that pair already depends on.

**Worth calling out explicitly, because it is the thing that could have made this fix silently useless:** you cite #320 as the same family, and that is the case where a non-popup screen stops the universe drawing, `DrawCompletedEvt` never fires, and the sim thread parks — which would leave the queued `ResetStatus` sitting in the queue until the screen closed, reproducing the very bug this fixes. `ShipListScreen` sets `IsPopup = true` (`ShipListScreen.cs:85`), so the universe keeps drawing and the sim keeps draining the queue. It works here for a reason that does *not* hold for its sibling screens, which is worth knowing before this pattern gets copied to one of them.

## Please change: the refresh walks a live UI list from the sim thread

```csharp
public void ResetStatus()
{
    foreach (ShipListScreenItem sel in ShipSL.AllEntries)
        sel.StatusText = ShipListScreenItem.GetStatusText(sel.Ship);
}
```

`ScrollList.AllEntries => Entries` (`ScrollList.cs:157`) returns the **live** list, not a snapshot. So the sim thread enumerates a collection the UI thread owns and mutates — on sort, on filter, on layout rebuild. If they collide it is an `InvalidOperationException` raised on the sim thread, which is the crash family we keep chasing in this codebase.

This is pre-existing on `:404`, so it is not something you introduced. But that line is behind shift-click mass-scrap, which is rare; this change puts the same walk on the ordinary scrap click, so the exposure goes up a lot.

The shape I would like instead flips a flag on the sim thread and lets the UI thread do its own walk:

```csharp
Screen.Universe.RunOnSimThread(() => Screen.StatusDirty = true);

// ShipListScreen.Update:
// if (StatusDirty) { StatusDirty = false; ResetStatus(); }
```

That keeps the semantics you are after — refresh only once the goal has actually run — while the list is only ever touched by the thread that owns it. It also lets you point `:404` at the same flag and retire the existing hazard in the same change. A `.ToArray()` snapshot would narrow the window but not close it, since copying a concurrently-mutated list can throw as well.

## Minor, pre-existing, take it or leave it

`:383-390` is an `if (!IsScuttle) ... else ...` whose two branches are byte-identical (`StatusText = GetStatusText(Ship);` either way), and the value is overwritten at `:411` regardless. Dead branching, not yours, but it is directly adjacent to your diff if you feel like removing it.

PR #353, opened 2026-07-22
Show total EMP tolerance in the shipyard design panel (#323)

### LINKED ISSUE #323: EMP protection inconsistency when loading a ship in the shipyard
When you load a ship in the shipyard, it shows an EMP protection in the short overview equal to the EMP protection of the modules + the size of the ships in slots. When you have actually loaded the ship, the EMP protection shown to the right of the ship is just the EMP protection of the modules.

Ludoal:
The shipyard design info panel displayed `BonusEMPProtection` (modules only) while the design load-list overlay shows `EmpTolerance` (hull surface area + modules) — the inconsistency described in the issue. The panel's own tooltip promises "the total EMP protection of this design".

Both now display the effective `EmpTolerance`, the value combat actually tests against (`EMPDisabled = EMPDamage > EmpTolerance`).

Fixes #323


gkapulis (approved):
Verified: EmpTolerance (SurfaceArea + BonusEMPProtection) is what combat tests (EMPDisabled = EMPDamage > EmpTolerance), the load-list overlay already shows it, and the TT_EmpProtection tooltip promises the total. ValNZ->Val is right since EmpTolerance is never zero. Merging - will strip the inline comment in a follow-up on the branch.

PR #354, opened 2026-07-22
Cancel a manual colonization when its colony ship is redirected by the player (#324)

### LINKED ISSUE #324: You can't cancel the colonisation of an unexplored planet, leading to subsequent colony ships going there (to their death)
I was in a fight with Remnant around a star system and thought that I had defeated them and sent in a colony ship to colonise an unexplored planet. But there were still a few Remnant so I sent the colony ship away again. But other colony ships that I owned kept trying to go to the planet as it probably still had the colonise command active. There is no way to cancel that for an unexplored planet. So I was constantly fighting the colonisation automation to stay away from that planet till I could defeat the remaining Remnant there. 

If you move away a colony ship that was en route to a planet to colonise it, then it should cancel the colonisation order for that planet. There is no good reason to move a colony ship away from the colonisation command other than that you don't want it to happen.

Ludoal:
When a player-ordered colonization's colony ship was manually sent elsewhere, `WaitForColonizationComplete` looped back and grabbed the next idle colony ship, so the automation kept feeding ships to the target — deadly for an unexplored planet guarded by Remnants, where no UI exists to cancel the order.

This hooks the actual player gesture: a direct, non-queued move or orbit order issued through `ShipMoveCommands` to a colony ship that is the assigned ship of a manual `MarkForColonization` retires that goal (a shift-queued waypoint is not treated as a change of mind). Ordering colonization of another planet was already handled by the goal constructor. AI and auto-colonize goals keep their retry behavior, and the ship itself is never touched — it is already following the new player order.

Note: an earlier attempt keyed on the absence of the ship's colonize plan misfired on legitimate transients (in-flight route replans) and mass-killed colonization goals; hooking the explicit input path avoids that whole class of false positives.

Fixes #324

### DIFF
--- a/Ship_Game/Universe/UniverseScreen/ShipMoveCommands.cs
+++ b/Ship_Game/Universe/UniverseScreen/ShipMoveCommands.cs
@@ -94,7 +94,11 @@ namespace Ship_Game.Universe
 
             // if ALT key is down, always Orbit the planet
             if (Input.IsAltKeyDown)
+            {
                 ship.OrderToOrbit(planet, clearOrders, GetStanceType());
+                if (clearOrders)
+                    CancelAbandonedColonizationGoal(ship);
+            }
             else if (ship.ShipData.IsColonyShip)
                 PlanetRightClickColonyShip(ship, planet, clearOrders); // This ship can colonize planets
             else if (ship.Carrier.AnyAssaultOpsAvailable)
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified: WaitForColonizationComplete really does re-order the redirected ship (loops back through CheckIfColonizationIsSafe -> OrderShipToColonize), RemoveGoal marshals to the sim thread from UI callers, the IsManualColonizationOrder + FinishedShip filter leaves AI/auto-colonize retry intact, and the colonize-elsewhere case is handled by the goal ctor as described. Merging - will strip the inline comments in a follow-up on the branch.

PR #355, opened 2026-07-22
SnapViewShip: set ShipToView so ship notifications zoom to the right ship

Ludoal:
`SnapViewShip` armed the camera follow flags (`snappingToShip` / `ViewingShip`) but never assigned `ShipToView`, and the camera update loop follows `ShipToView` — so clicking a "Research/Mining Station built" notification (or any other `SnapToShip` notification) zoomed to whatever ship was viewed last instead of the referenced ship. Mostly latent while `ShipToView` stayed null; any feature that sets it makes the wrong-zoom reproducible.

Assign the target ship, matching what `ViewToShip` already does.


gkapulis (approved):
Verified against the camera code: ShipToView is the only thing AdjustCamera follows, it survives deselection, and SnapViewShip's flags were dead without it. Single caller (NotificationManager SnapToShip). Merging - will strip the inline comment in a follow-up on the branch.

PR #356, opened 2026-07-22
Refresh clickable build goals every frame: paused DSB orders show on the map (#285)

### LINKED ISSUE #285: Platforms ordered via the deep space building menu are not shown on the map while the game is paused
Something changed since the last update. You used to immediately get a symbol on the map when you placed any form of platform, shipyard or station on the map via the deep space building menu. Now that is not the case anymore. You only get to see the symbol when you unpause the game. It is confusing as you wonder whether you misclicked. It also means that it is harder to place a bunch of these platforms in a relative position to one another as you can't see the ones that you already placed.

Note: it also used to be the case that you couldn't see the ordered platforms after loading a game and before unpausing. But that was strictly limited to the moment that you loaded a game and only related to the platforms that were ordered before you had saved and reloaded the game. When you started placing additional platforms, the game would immediately update, paused or not, and show symbols for all the platforms that were ordered before. Now, the issue is that you can't see platforms at any time that you place platforms while the game is paused.

Ludoal:
Platforms ordered via the deep space building menu were not shown on the map while the game was paused (a regression the report notes — they used to appear instantly), and DSB icons under construction could not be selected in pause either.

Cause: `UpdateClickableItems` only ran from the simulation path, which does not advance while paused, and the icons' screen positions are camera-dependent so they also went stale on any camera move in pause. The fix refreshes the clickable build goals from the screen update as well — the scan covers player deployment goals only, so the per-frame cost is trivial.

This fix has been running in the Ludoal fork since Patch 45 builds (it was originally made for the selection-in-pause half of the symptom) with no observed side effects.

Fixes #285

### DIFF
--- a/Ship_Game/Universe/UniverseScreen/UniverseScreen.cs
+++ b/Ship_Game/Universe/UniverseScreen/UniverseScreen.cs
@@ -701,6 +701,12 @@ namespace Ship_Game
             {
                 SystemInfoOverlay.Update(elapsed);
             }
+            // The clickable build goals froze while paused (their screen positions are
+            // camera-dependent and the sim loop owns the refresh): platforms ordered via
+            // the deep space building menu were invisible on the map until unpausing, and
+            // selecting a DSB under construction was impossible in pause. Refreshing here
+            // too is trivially cheap (player deployment goals only).
+            UpdateClickableItems();
             if (ShowPlanetInfo)
             {
                 pInfoUI.Update(elapsed);


gkapulis:
Thanks for the diagnosis — it's exactly right on both symptom halves (invisible new DSB goals in pause + unselectable under-construction icons, both because the refresh only ran from the unpaused sim path and the projections are camera-dependent).

We applied the fix ourselves in a different place though (commit d804bdd19 on `jupiter-1.60/fixes_28`), so closing this PR: calling `UpdateClickableItems()` from `UniverseScreen.Update` runs it on the **UI thread**, and it reads `Player.AI.Goals` — `GoalsList` is owned by the simulation thread (see the contract comment above `EmpireAI.MarshalScreen`). We had a Sentry-reported torn-read crash from exactly this cross-thread read and fixed it in 1.60 by marshaling all off-thread mutations to the sim thread; a UI-thread reader would quietly reopen that race (the `// ToArray() used for thread safety` comment in that method is unfortunately not a real guarantee — `Array<T>.ToArr()` isn't atomic against a concurrent resize). It also double-runs against the existing sim-path call whenever the game is unpaused.

The applied version calls `UpdateClickableItems()` from the **paused branch of `ProcessSimulationTurns`**, right after `InvokePendingSimThreadActions()` — that branch runs once per draw frame on the sim thread even while paused, so new DSB orders become visible/selectable the same frame they land in `GoalsList`, camera moves in pause refresh the projections, and the list never leaves its owning thread. Credit noted in the commit body.

PR #357, opened 2026-07-22
Diplomacy: tech offer count matches the list actually shown (#342)

### LINKED ISSUE #342: Trading Technologies
I want to trade my technologies to an AI. Negotation UI has "Technology... (1)", but when I click it, it turns yellow, but no technologies are listed. I can view the AI's technologies, and they number corresponds to the number that drop down in the list. But not for my side of the negotiation table.

Ludoal:
The Technology header in the negotiation panel counted all tradable techs, but the sublist then skipped every tech the receiving AI cannot use (no pre-made designs) — so the player could see "Technology (1)" expand to an empty list, while the AI's own side always displayed correctly. Filter first, count after; the design rule itself (FB: don't trade ship tech the AI can't use) is unchanged.

Fixes #342

### DIFF
--- a/Ship_Game/GameScreens/DiplomacyScreen/DiplomacyOffersComponent.cs
+++ b/Ship_Game/GameScreens/DiplomacyScreen/DiplomacyOffersComponent.cs
@@ -104,16 +104,21 @@ namespace Ship_Game.GameScreens.DiplomacyScreen
             var theirDesigns = Them.AllFactionShipDesigns;
             
             Us.AI.TradableTechs(Them, out Array<TechEntry> tradeAbleTechs);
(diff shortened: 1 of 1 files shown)


gkapulis:
Applied directly to jupiter-1.60/fixes_27 (credited in the commit body), same reason as #358/#359 — stale-fork PRs vs the branch-protection up-to-date rule. Verified: pure De Morgan inversion of the skip condition (identical set, WeCanUseThisTech is side-effect-free, and the explicit Them.isPlayer short-circuit preserves the old IsHidden-skip behavior for the player exactly). Explains the issue's asymmetry too: Them is the AI only on the player's panel. Thanks!

PR #358, opened 2026-07-22
Loading screen: Click to Continue no longer overlaps the advice text

Ludoal:
On the load-game screen, "Click to Continue!" overlapped the advice/hint text: the vertical offset was derived from the twin block in `CreatingNewGameScreen` but with the sign of the `LineSpacing` term flipped the wrong way. One-character fix, reported on the fork from a screenshot.


gkapulis:
Applied directly to jupiter-1.60/fixes_27 (credited in the commit body) — verified against the CreatingNewGameScreen twin block, and while in there also switched the label to the localized GameText.ClickToContinue token like the twin uses. Applying by hand since the branch-protection up-to-date requirement makes stale-forked PRs a pain to merge (same as #359). Thanks!

PR #359, opened 2026-07-22
Fix NullRef in Ships Array status text when no solar system is reachable

Ludoal:
`ShipListScreenItem.GetStatusText` calls `FindClosestSystem` and dereferences the result; in universes where the relevant ship has no reachable system this threw a NullReferenceException while drawing the Ships Array. Found via a draw-loop crash log on the fork's battle-simulator arena (a minimal universe), but the guard is valid for any universe where FindClosestSystem legitimately returns null.

### DIFF
--- a/Ship_Game/ShipListScreenItem.cs
+++ b/Ship_Game/ShipListScreenItem.cs
@@ -298,8 +298,10 @@ namespace Ship_Game
 
                         if (!ship.AI.OrderQueue.TryPeekLast(out ShipAI.ShipGoal last))
                         {
+                            // Ludoal fork: FindClosestSystem is null in a system-less
+                            // universe (battle simulator arena) — NullRef froze the UI
                             SolarSystem system = ship.Universe.FindClosestSystem(ship.AI.MovePosition);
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified: FindClosestTo(systems, count, to) returns null when count == 0 (CollectionFindClosestTo.cs), so FindClosestSystem legitimately nulls in a system-less universe. These two GetStatusText derefs were the only unguarded consumers, and the ExploringTheGalaxy fallback is already the right behavior for both. Zero mainline exposure (generation always creates systems) but the guard is free and hardens a draw path. Pushed a small commit to the branch removing the inline comment (history lives here in the PR).

gkapulis:
Applied directly to jupiter-1.60/fixes_27 as 5c3e6f903 (credited to you in the commit body) — GitHub's PR-sync kept lagging behind the branch-protection up-to-date requirement after we updated your branch, so landing it by hand was faster than fighting the merge button. The fix itself is verified and unchanged: both guards, with the existing ExploringTheGalaxy fallback. Thanks for the catch — the guard is valid beyond the battle-sim arena, for any universe where FindClosestSystem can return null.

PR #360, opened 2026-07-22
A UI draw failure must not starve the simulation thread

Ludoal:
`UniverseScreen.Draw` signals `DrawCompletedEvt` at the end of the frame — it is what lets the simulation thread run its next turn. An exception anywhere in `DrawGeneralUI` skipped that signal, freezing the entire simulation for as long as the failing UI element kept drawing, with no diagnostic. We hit this on the fork (a UI element threw every frame: game appeared hard-frozen); with the guard, the failure logs once with a stack trace and the sim thread stays alive.

The catch is deliberately broad and logs only the first occurrence: a UI draw failure should degrade the UI, never starve the simulation.

### DIFF
--- a/Ship_Game/Universe/UniverseScreen/UniverseScreen.Draw.cs
+++ b/Ship_Game/Universe/UniverseScreen/UniverseScreen.Draw.cs
@@ -440,7 +440,22 @@ namespace Ship_Game
                 DrawShipAndPlanetIcons(batch);
                 DrawSolarSystems(batch);
                 DrawSystemThreatIndicators(batch);
-                DrawGeneralUI(batch, elapsed);
+                // Ludoal fork: a UI draw failure must never starve the sim thread —
+                // an exception here used to skip DrawCompletedEvt.Set() below, freezing
+                // the simulation for as long as the failing element kept drawing
+                // (battle sim arena: selecting a ship froze the game). Log and go on.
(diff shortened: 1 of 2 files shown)


gkapulis (changes requested):
Verified the diagnosis against the draw/sim handshake and it's exactly right: an exception escaping `DrawGeneralUI` is caught by `ScreenManager.Draw`, which logs and keeps looping, so `DrawCompletedEvt.Set()` never runs and `ProcessTurns` blocks on `WaitOne()` forever — silent hard freeze. Good catch, and the guard is the right shape.

Two changes before we can take it:

1. **Don't swallow `ObjectDisposedException`.** `ScreenManager.Draw` has a dedicated `catch (ObjectDisposedException)` that triggers the reload-all-screens recovery when the device sleeps and graphics resources are lost. A broad `catch (Exception)` here would eat that ODE, suppress the UI permanently, and the recovery never fires. Please add before the generic catch:

   ```csharp
   catch (ObjectDisposedException)
   {
       throw; // let ScreenManager's device-lost recovery handle it
   }
   ```

2. **Cosmetics:** drop the `// Ludoal fork:` comments in both files (the history lives in this PR), and use `Exception` instead of `System.Exception` — the file already has `using System;`.

With those, happy to merge.

PR #361, opened 2026-07-22
Camera chase survives selection: stop clearing ViewingShip every frame

Ludoal:
`UpdateSelectedShips` refreshes the selected ship's UI every frame via `SetSelectedShip`, which unconditionally called `ClearSelectedItems` with `clearFlags: true` — so `ViewingShip`/`snappingToShip` were switched off EVERY frame while a ship was selected. Since chasing a ship implies having it selected, the chase camera could never survive its initial snap; the middle-click follow feature has effectively been broken by this for a long time.

`SetSelectedShip` gains a `clearFlags` pass-through (default `true`, so every other call site keeps its exact behavior); the per-frame refresh passes `false`. Field-tested on the Ludoal fork since Patch 45 builds — follow now works and disengages normally on player camera input.

### DIFF
--- a/Ship_Game/Universe/UniverseScreen/UniverseScreen.SetSelectedItem.cs
+++ b/Ship_Game/Universe/UniverseScreen/UniverseScreen.SetSelectedItem.cs
@@ -58,7 +58,7 @@ public partial class UniverseScreen
         int num = SelectedShipList.Count();
         SelectedShipList.RemoveInActiveObjects();
         if (SelectedShip != null)
-            SetSelectedShip(SelectedShip, SelectedFleet);
+            SetSelectedShip(SelectedShip, SelectedFleet, clearFlags: false); // same ship, UI refresh only - keep the chase camera alive
         else if (num != SelectedShipList.Count())
             SetSelectedShipList(SelectedShipList, SelectedFleet);
     }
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified: UpdateSelectedShips runs every frame from HandleInput:312 and SetSelectedShip -> ClearSelectedItems(clearFlags:true) wiped ViewingShip/snappingToShip/returnToShip while any ship was selected, which is exactly the chase precondition. The clearFlags default keeps all seven other call sites byte-identical, and every disengage path (middle-drag pan, edge panning, middle-click toggle, minimap click, ship-panel double-click) still clears the flags.

Worth noting it also revives returnToShip: the 'return to your ship' paths from the colony screen (HandleInput:189-195) and ShipListScreen were collateral damage of the same wipe. Merging - follow-up on the branch will null ShipToView when the followed ship is removed, so the camera does not lock onto a dead ship (AdjustCamera already has the 'ShipToView == null -> ViewingShip = false' guard waiting for it).

gkapulis:
Merged, with follow-up b510d2ee6 on jupiter-1.60/fixes_29: ShipToView is now nulled when the followed ship is removed (feeding AdjustCamera's existing 'ShipToView == null -> ViewingShip = false' guard), so a chase releases cleanly when the target dies instead of locking the camera on its last position. Also tracked CamDestination to the chased ship, so ending a chase for any reason continues from where the camera is rather than gliding back to where the chase started. Comments stripped per our convention.

Good find - worth knowing it also revived returnToShip, so 'return to your ship' after closing a colony screen or the ship list works again too.

PR #362, opened 2026-07-22
Patrols screen closes with its own hotkey (P), not L

Ludoal:
The Patrols screen's close handler listened for `Keys.L` — a copy-paste from PlanetListScreen that was never adapted. Pressing P (the key that opens the screen) over the open Patrols screen appeared to do nothing, while L closed it. It now uses the same input binding that opens it. Field-tested on the Ludoal fork since Patch 45 builds.


gkapulis:
Thanks,  committed into fixes branch

PR #364, opened 2026-07-22
Starving planets no longer export their food buffer

### LINKED ISSUE #314: Governor exporting all the food and production from a planet that can't support itself
When you just colonise a planet with and advanced colony module, you get a planet with 20/20 food and 20/20 production. The industrial governor on a barren planet sets both on export immediately because the tiny storages are full. When you have a thriving empire, it will also immediately start importing population to the planet. The result is that one freighter exports all 20 food from the planet and other freighters bring so much population that the planet quickly goes into negative food per turn and starts starving.

My suggestion in general: 

1. A planet that is not a serious food producer should not export all its food. It should leave enough food that with a small population growth, it can still feed itself. Ideally, in this case of a just founded planet, you just leave the 20 food there. The +0.1 food per turn from a just founded planet is not a big surplus.
2. A planet that is building stuff should not export all its production. It should leave enough for several turns of infrastructure. Even if a planet is supporting the empire, you generally don't need it to go fully empty. Ideally, in this case of a just founded planet, the 20 production are left there to help start the planet.

In general, I see planets receiving a full production or food supply and then just by having this full supply (like 180/180 for 5 seconds before using some of the food again) go into export modus. I guess that when one planet is full of food that it just received and a neighbour is about to starve, then it is better to transfer some of it. But if a planet is reliant on food (or production) imports to keep alive (or keep production going at full speed) and another can just freely export, then the one that can freely export and doesn't need it for itself should be given preference to export.

Ludoal:
Fixes #314.

Three-stage fix for the export-while-starving window:
- DetermineFoodState: negative max food production now always resolves
  to STORE (previously only above 0.9 storage ratio, so a declining
  planet in the [export threshold, 0.9] window kept exporting while
  melting toward famine).
- GetFoodExportSlots: the guaranteed minimum export slot drops to 0
  when net food income is negative.
- The Agricultural exemption from the import threshold no longer
  applies when the planet cannot feed itself even at max farming
  (crushed fertility): the ruined farm finally accepts relief.

Test status: compiled and logic-reviewed against the issue's repro; play-testing on our fork is partial.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet_Resources.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet_Resources.cs
@@ -161,12 +161,12 @@ namespace Ship_Game
 
             float ratio               = Storage.FoodRatio;
             bool belowImportThreshold = ratio < importThreshold 
-                                        && CType != ColonyType.Agricultural
+                                        && (CType != ColonyType.Agricultural || Food.NetMaxPotential < 0)
                                         && Food.NetFlatBonus < Consumption;
 
             // This will allow a buffer for import / export, so they dont constantly switch between them
             if      (ShortOnFood() || belowImportThreshold)   FS = GoodState.IMPORT; 
-            else if (Food.NetMaxPotential < 0 && ratio > 0.9) FS = GoodState.STORE;  // We are negative on food production but have a lot of food
+            else if (Food.NetMaxPotential < 0)                FS = GoodState.STORE;  // Negative food production: keep the buffer, never export it away
             else if (ratio > exportThreshold)                 FS = GoodState.EXPORT; // Until we get back to the Threshold, then export
             else                                              FS = GoodState.STORE;  // We are between our thresholds
         }
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet_Trade.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet_Trade.cs
@@ -121,7 +121,8 @@ namespace Ship_Game
             if (ManualFoodExportSlots > 0 && Owner == Universe.Player)
                 return ManualFoodExportSlots;
 
-            int min = Storage.FoodRatio > 0.75f ? 2 : 1;
+            // a planet LOSING food guarantees no export slot - the last freighter no longer leaves no matter what
+            int min = Food.NetIncome < 0 ? 0 : Storage.FoodRatio > 0.75f ? 2 : 1;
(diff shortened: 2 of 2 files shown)


gkapulis:
Closing this one — we took #314 in-house and it landed on `jupiter-1.60/fixes_30` in 146948b2b. Not a reflection on the work here; two of your findings went straight into it and are credited in the commit.

**What we kept from your diagnosis:**

- The forced-minimum export slot. `Clamped(v, min, max)` is `Max(min, Min(v, max))`, so with a negative `NetIncome` the old `min` of 1-or-2 pushed at least one freighter out of a planet actively losing food. Your "the last freighter no longer leaves no matter what" was literally accurate, and that behaviour is gone.
- The layering. You had spotted that the state machine and the slot count bite in different windows — `NetMaxPotential >= 0` but `NetIncome < 0` is a real, distinct case, and it survives in the final shape.

We also took the `DetermineFoodState` change as-is: a planet that cannot feed itself even at maximum farming now always STOREs, not only above 0.9 storage.

**Why we rebuilt rather than merged:**

The reporter's own repro was not covered. His planet has 20/20 food and **+0.1 food per turn** — income is *positive*, so neither `NetMaxPotential < 0` nor `NetIncome < 0` fires. It still falls through to `ratio > exportThreshold` (Industrial is 0.99, ratio is 1.0) and still gets `min = 2` from `FoodRatio > 0.75`. Two freighters still drain all 20 food, exactly as reported. The guards only engage once the imported colonists push it negative, by which point the buffer is already gone — so the change treated the decline phase rather than the cause.

The fix needed a reserve floor instead of a negative-income gate, sized against the population the colony is growing into rather than today's tiny headcount, because a young colony consumes almost nothing right up until it fills out.

Two things surfaced while building it that are worth knowing for next time, because neither is visible from the diff:

- **Gating export slots is advisory, not a guarantee.** How much a freighter actually lifts comes from `ExportGoodsLimit`, which caps at `FoodHere` — the entire stock — and knows nothing about slot counts. A single slot granted on expected income still drains everything, and where several freighters share a planet the per-ship limit divides *total storage* rather than the surplus. The reserve had to bind there too.
- **Issue #314 has two numbered asks**, and the second is production: "a planet that is building stuff should not export all its production". `GetProdExportSlots` carries the same forced-minimum four lines below the one you changed and was untouched here, so "Fixes #314" was only ever half true. The in-house fix reserves against the colony's own build queue as well.

Thanks for the work on this — the negative-income window was a real bug and it would not have been found from the issue text alone.

PR #365, opened 2026-07-22
Lighter-materials research no longer weakens mass-reduction devices

### LINKED ISSUE #341: Inertial damper becomes less effective when you research technology that creates lighter materials
The inertial damper reduces the weight of a ship by a certain amount. The base 2x2 inertial damper reduces the weight of a ship by 100 when powered. However, when you research improved lighter materials -x%, then the inertial damper also reduces the weight of the ship by x% less. So for instance, when you research improved lighter materials -20%, then the 2x2 inertial damper will only reduce the weight of ships by 80 instead of 100. The inertial damper technology becomes less effective when we research lighter materials. 

I guess this is just a simple code mistake where the lighter materials logic is applied on the total ship weight. But it should only be applied on the sum of positive weights of a ship and not the negative weight of the inertial dampers.

Ludoal:
Fixes #341.

InitializeMass applied the empire MassModifier to the signed SUM of
module masses, so a -20% materials tech also scaled the negative mass
of inertial dampers (a 100 reduction became 80). The per-module path
(GetActualMass with useMassModifier) already applies the modifier to
positive mass only; the ship-total path now does the same: hull and
modules get lighter, reduction devices keep their full effect.

### DIFF
--- a/Ship_Game/Ships/ShipStats.cs
+++ b/Ship_Game/Ships/ShipStats.cs
@@ -162,12 +162,20 @@ namespace Ship_Game.Ships
         public float InitializeMass(ShipModule[] modules, Empire loyalty, int surfaceArea, float ordnancePercent)
         {
             int minMass = (int)(surfaceArea * 0.5f * (1 + surfaceArea * 0.002f));
-            float mass = minMass;
+            float positiveMass = minMass;
+            float negativeMass = 0f;
 
             for (int i = 0; i < modules.Length; i++)
-                mass += modules[i].GetActualMass(loyalty, ordnancePercent, useMassModifier: false);
+            {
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified against the source. InitializeMass is the single runtime mass path (ShipStatusChange -> UpdateMassRelated), and it scaled the signed sum; the per-module display path (ShipModule.GetActualMass:1231-1237) already applies MassModifier to positive mass only and returns raw negative for powered reduction devices — so the module tooltip promised -100 while the ship total delivered -80. This fix aligns the total with the tooltip. Also confirmed no drift: unpowered-damper penalties scaled before and still do, minMass handling and the floor are unchanged. Pre-existing nit (not for this PR): GetMass(Empire) cross-empire ratio conversion is now slightly approximate for damper ships — display/comparison only.

PR #366, opened 2026-07-22
Subspace Tunneling FTL bonus actually applies

### LINKED ISSUE #328: The subspace tunneling technology doesn't give its promised 30% increase in warp speed
Image

Image

[Jupiter__1.60.00043_Combined_Arms-v1.60.0005_The Kulrathi Shogunate, Star Date 1176.3.zip](https://github.com/user-attachments/files/29159818/Jupiter__1.60.00043_Combined_Arms-v1.60.0005_The.Kulrathi.Shogunate.Star.Date.1176.3.zip)

In this game, the Pollops discovered the Subspace Tunneling technology on some planet with some event. I traded for it, but I am not getting the technology bonus. It should give me an additional 30% warp speed, but I am left with the 100% warp speed increase that I already got from technologies, not 130%. I did not only check the values shown in the diplomacy screen, but also the values shown of the actual ships in the ship array and it confirms that it is just a 100% increase in speed.

I myself found the Remnant Assembly technology and could see the stats of some weapons increase according to what it says that it does. I am not sure whether the Subspace Tunneling technology doesn't work at all or whether it has to do with trading for the technology. I recall that in a game that I played years ago, it also didn't work, so it's not a new bug.

Ludoal:
Fixes #328.

The Slip Streams secret tech grants a bonus named 'Subspace Tunneling',
but the unlock switch only had the generic 'FTL Speed Bonus' label -
the flavor alias was never wired (its siblings all have both). The +30%
warp bonus silently did nothing, researched or traded. One case label.

### DIFF
--- a/Ship_Game/TechEntry.cs
+++ b/Ship_Game/TechEntry.cs
@@ -959,9 +959,9 @@ namespace Ship_Game
                 case "Reactive Armor":
                 case "Armor Explosion Reduction": data.ExplosiveRadiusReduction += unlockedBonus.Bonus; break;
                 case "Slipstreams":
+                case "Subspace Tunneling": // flavor name used by the Slip Streams secret tech - it had no case, so its +30% did nothing (issue 328); the tech text promises an IN-BORDERS bonus, not a global one
                 case "In Borders FTL Bonus": data.Traits.InBordersSpeedBonus += unlockedBonus.Bonus; break;
                 case "StarDrive Enhancement":
-                case "Subspace Tunneling": // the Slip Streams secret tech uses this flavor name - it had no case, so its +30% did nothing (issue 328)
                 case "FTL Speed Bonus": data.FTLModifier += unlockedBonus.Bonus * data.FTLModifier; break;
                 case "FTL Efficiency":
                 case "FTL Efficiency Bonus": data.FTLPowerDrainModifier -= unlockedBonus.Bonus * data.FTLPowerDrainModifier; break;
--- a/Ship_Game/TechEntry.cs
+++ b/Ship_Game/TechEntry.cs
@@ -961,6 +961,7 @@ namespace Ship_Game
                 case "Slipstreams":
                 case "In Borders FTL Bonus": data.Traits.InBordersSpeedBonus += unlockedBonus.Bonus; break;
                 case "StarDrive Enhancement":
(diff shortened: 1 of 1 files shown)


gkapulis (changes requested):
The diagnosis is exactly right — the switch matches `BonusType ?? Name`, the Slip Streams XML (vanilla and Combined Arms both) carries only `<Name>Subspace Tunneling</Name>`, there's no matching case and no default, so the bonus silently evaporates. Confirmed against the source.

But the routing needs to move one group up. The player-facing bonus text (GameText 18272, identical in vanilla and CA) reads: "Ships traveling within your own empire's **borders** gain 30% bonus to their FTL speeds." — that's `InBordersSpeedBonus` (applied only under friendly influence, Ship_Movement.cs:108), the same stat the adjacent `Slipstreams` flavor alias feeds. As written, the PR wires it to the global `FTLModifier`, which grants +30% warp everywhere — stronger than the tech advertises.

Suggested fix:

```csharp
case "Slipstreams":
case "Subspace Tunneling": // flavor name used by the Slip Streams secret tech (issue 328)
case "In Borders FTL Bonus": data.Traits.InBordersSpeedBonus += unlockedBonus.Bonus; break;
```

Two side notes: (1) when retesting, the number only moves while ships sit inside your own influence — the diplomacy-screen FTL% and open-space ship stats won't change, which is also worth mentioning to the issue reporter; (2) the CA XML's inline `<Description>` says 40% while the localized text and `Bonus` value say 30% — harmless (BonusIndex text is what displays) but worth a content cleanup someday.

Ludoal:
Rerouted to InBordersSpeedBonus as suggested — the case now sits with the Slipstreams alias where it belonged. Comment carries the in-borders rationale so the next reader doesn't rewire it back to the global modifier. Will pass the retest note (bonus only visible inside own influence) to the issue reporter, and noted the CA 40%-vs-30% description mismatch for a content cleanup. Thanks for the catch on the routing.

gkapulis (approved):
Revision verified: routed to the InBordersSpeedBonus group as requested - matching what GameText 18272 (the tech's own BonusIndex) promises, and consistent with how Ship_Movement consumes it in friendly influence. The case string matches Slip Streams.xml's <Name> exactly and appears in no other tech XML. Merging - will trim the inline comment in a follow-up on the branch. Reminder for the issue-328 reporter: the number only moves inside your own influence border.

PR #367, opened 2026-07-22
Stations need 10% hangar area for the Dedicated Carrier design note

### LINKED ISSUE #317: A platform with less than 10% modules dedicated to hangars is seen as a dedicated carrier
Image

I thought that the rule was that you needed more than 10% of the station/platform/ship dedicated to hangars to be considered a dedicated carrier. But in this case this station with 24 out of 356 slots dedicated to hangars is considered a dedicated carrier. Actually, even with a single hangar of 12 slots, it is considered a dedicated carrier. Are the rules different for a station or is there some relation to it being a research station?

Ludoal:
Fixes #317.

CheckDedicatedCarrier fired for ANY stationary design with any fighter
hangar (a 24-slot bay on a 356-slot station, or a single 12-slot
hangar). Stationary designs are now held to the same 10% hangar-area
threshold that ships need to classify as carriers. Informative-level
issue, display only.

### DIFF
--- a/Ship_Game/GameScreens/ShipDesign/ShipDesignIssues.cs
+++ b/Ship_Game/GameScreens/ShipDesign/ShipDesignIssues.cs
@@ -508,9 +508,12 @@ namespace Ship_Game.GameScreens.ShipDesign
         }
 
         public void CheckDedicatedCarrier(bool hasFighterHangars, RoleName role, 
-                                          int maxWeaponRange, float sensorRange, bool shortRange)
+                                          int maxWeaponRange, float sensorRange, bool shortRange,
+                                          float hangarAreaPercent = 1f)
         {
-            if (role != RoleName.carrier  && !Stationary || !hasFighterHangars)
+            // stationary designs were flagged 'Dedicated Carrier' from a single hangar bay;
+            // hold them to the same 10% hangar-area threshold ships need for the carrier role
+            if (role != RoleName.carrier && !(Stationary && hangarAreaPercent > 0.1f) || !hasFighterHangars)
                 return;
 
             bool minCarrier  = false;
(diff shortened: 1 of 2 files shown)


gkapulis (changes requested):
Right problem, right intent, and the issue's repro does get fixed — but the value being thresholded is not the one the description claims, and it will misfire on a different class of station.

## The guard change is correct

Worth confirming the diagnosis since the precedence is easy to misread. The original parses as `(role != carrier && !Stationary) || !hasFighterHangars`, i.e. show the note whenever `hasFighterHangars && (role == carrier || Stationary)`. So yes — **any** stationary design with **any** hangar was flagged. The new form gates the stationary arm behind the percentage, and the logic is sound.

## But it is measuring a different quantity

The description says stations are now "held to the same 10% hangar-area threshold that ships need to classify as carriers." They are not. The two values are unrelated:

| | expression | what it measures |
|---|---|---|
| Role classifier (`RoleData.cs:69`, via `SurfaceAreaPercentOf` at `:140-143`) | `Modules.SurfaceArea(m => Hangar && !IsSupplyBay && !IsTroopBay) / SurfaceArea` | the hangar modules' **footprint on the hull** — `ShipExtensions.SurfaceArea` sums `m.XSize * m.YSize` |
| This PR | `ds.TotalHangarArea / S.SurfaceArea` | sum of `MaximumHangarShipSize` — the largest **fighter each bay may launch** |

`MaximumHangarShipSize` is a `short` on `ShipModuleFlyweight` describing what a bay may *contain*; it is what `FighterScrollList` uses to decide which fighters fit. It says nothing about how much hull the module occupies, and a physically compact hangar can permit a large fighter.

**The filter differs as well.** The role rule explicitly excludes supply bays and troop bays, precisely so that a supply or troop platform does not read as a carrier. `TotalHangarArea` is `modules.Sum(m => m.MaximumHangarShipSize)` (`ShipDesignStats.cs:111`) with no filter at all. So a station packed with troop bays can clear the new 10% gate and still receive the Dedicated Carrier note — which is the original complaint in different clothing.

The issue's own repro passes because of scale rather than rule: a single 24-size hangar on a 356-surface station gives 0.067, comfortably under the line. That is luck, not the stated threshold.

## What I would like instead

Compute the same expression the classifier uses, so the note and the role agree by construction rather than by coincidence. Ideally by exposing `RoleData`'s check (or a shared constant) rather than duplicating the predicate and the literal — `0.1f` would otherwise gain a third home in the codebase and the two copies will drift.

## Two smaller things

- `hangarAreaPercent = 1f` has exactly one caller, and the default silently means "always passes" for any future one. Please make it required rather than defaulted.
- With the percentage folded in, the condition is getting hard to read at a glance. Splitting the stationary case onto its own line would help the next person who has to parse the `&&`/`||` precedence the way I just did.

PR #368, opened 2026-07-22
Double refit keeps the ship's fleet affiliation

### LINKED ISSUE #295: When you double refit ships, they lose their affiliation to a fleet
When you refit ships of a fleet to say 'design A", they fly to a nearby planet to be improved. At that moment, they seemingly don't belong to their fleet anymore, but when they arrive at a planet, you can see that their link to their fleet is still there as they are still shown in the building queue as belonging to that fleet. And once the refit has finished, they become part of the fleet again and join with it at a central location.

When you recognise that you made a mistake and ordered them to be refit to 'design A' while you had wanted 'design B', you can still select them on their way to their refit location and give them a new refit command to design B. You can even select all ships of their type to be refitted to 'design B'. But now that they have received a second refit command, they will lose their affiliation to their fleet. In the building queue, they won't be shown as part of that fleet. And when refitted, they won't returns to their fleet.

Ludoal:
Fixes #295.

The first refit detaches the ship from its fleet (the link then lives
only in the refit goal's Fleet + the fleet node). A second refit order
read oldShip.Fleet - already null - then removed the old goal, so the
affiliation died with it and the rebuilt ship never rejoined its fleet.
The new goal now inherits the Fleet from the goal it replaces.

### DIFF
--- a/Ship_Game/Commands/Goals/RefitShip.cs
+++ b/Ship_Game/Commands/Goals/RefitShip.cs
@@ -45,7 +45,14 @@ namespace Ship_Game.Commands.Goals  // Created by Fat Bastard
                 VanityName = oldShip.VanityName;
 
             if (OldShip.AI.State == AIState.Refit)
+            {
+                // A ship already refitting was detached from its fleet by the FIRST refit goal
+                // (the fleet link lives in that goal, not on the ship anymore). Inherit the
+                // affiliation before removing the old goal, or a double refit loses the fleet.
+                if (Fleet == null && Owner.AI.FindGoal(g => g.Type == GoalType.Refit && g.OldShip == OldShip) is FleetGoal prior)
+                    Fleet = prior.Fleet;
                 RemoveOldRefitGoal();
+            }
         }
 
         GoalStep FindShipAndPlanetToRefit()


gkapulis:
Applied to jupiter-1.60/fixes_27 (root-cause credit to you in the commit body) with one necessary extension: inheriting the prior goal's Fleet is not sufficient on its own. The first goal's ClearFleet nulls node.Ship (Fleet.RemoveFromAllSquads), so the second goal's step 1 FindShipNode(OldShip) can never re-claim the node, and removing the prior goal leaves node.Goal pointing at a dead goal — AddShipDataAndFleet re-adds via FindNodeWithGoal(this), which then never matches. So the rebuilt ship still would not rejoin with the Fleet-only inherit. The landed version also takes over the prior goal's fleet node in the ctor (AssignGoal + AssignShipName before RemoveOldRefitGoal), which also makes triple+ refits chain correctly. Also folded in an adjacent same-class fix: BuildNewShip's planet-owner-changed failure path was stranding node.Goal by skipping RemoveGoalFromFleet. Your diagnosis of the detach mechanism was exactly right — it pointed straight at the node lifecycle. Thanks!

Issue #321, opened 2026-06-16
The construction of biospheres to increase the population should be independent of the tax rate


Roland-Johansen:
I noticed that when I raise my tax rate, then the governors start building biospheres, while if I lower it, then they scrap them again. This also happens automatically when you use a budget and let that balance your tax rate. I guess that is also happening for the AI. That doesn't make sense and hurts the economy of the player and AI, especially since the tax rates go up and down a bit during periods of ship building which means that the biospheres get constructed during the periods of high tax rate and deconstructed again when the tax rate drops after a period of ship building.

If a biosphere costs 0.2 and with a tax rate of 25% the say 200 million people living under the biosphere pay 0.2 in taxes (due to taxation and credit buildings), then that means that 75% of those people can do other stuff like production or science. 50 million pay taxes and 150 million do other stuff. It doesn't matter whether the actual empire wide tax rate is 15% or 50%, that biosphere is worth it as only 25% of the people living there are paying taxes can make it break even and the other 75% can do other stuff. 

To put it in numbers and not percentages. Say you have an empire of 1 billion people and a 100% tax bonus from buildings and 1 extra credit per person from buildings. Your costs are zero, so you need a zero tax rate. Now you build a biosphere that adds 200 million people and costs 0.2 maintenance. If you let 50 million people pay taxes, then you get 0.05 * (1+100% of 1) * (1+1)=0.2 income to pay the biospheres. Or to put it differently, you create a tax rate so that 0.05 billion (50 million) out of your now 1.2 billion people pay taxes. That is a tax rate of 0.05/1.2 = 4.1666% applied on your population of 1.2 billion, to get a tax income of 4.1666% of 1.2 * (1+ 100% of 1) * (1+1) =0.2. The other 95.8333% of your 1.2 billion people = 1.15 billion people can now work. So you increase the working population from 1 to 1.15 billion while your tax rate went up from 0% to 4.1666%. It would make no sense to delete that biosphere again when the tax rate drops a bit because some other planet got a higher tax income from some buildings or because some expenses were lowered. The biosphere is a positive, no matter what the tax rate is.

It doesn't matter what the tax rate was before and after, this biosphere is worth it because 0.05 out of the 0.2 billion people living there can already pay the maintenance.

gkapulis:
Fixed in **933196b4f** on `jupiter-1.60/fixes_33`, with a follow-up in **a5c45f34b**. It will ship in the next Jupiter patch.

[...]

PR #369, opened 2026-07-22
Governor keeps unprofitable biospheres the colony budget pays for

### LINKED ISSUE #313: The governor logic deletes my biospheres even if I gave it a huge budget like 99
[Mars__1.51.15106_Combined_Arms-v8.7i_Draylok biospheres, Star Date 1113.1.zip](https://github.com/user-attachments/files/28934962/Mars__1.51.15106_Combined_Arms-v8.7i_Draylok.biospheres.Star.Date.1113.1.zip)

I have manually ordered a bunch of biospheres to be constructed on Erfolk II to increase the population. The governor first constructs them all and then deletes them all again. The governor is following a plan in this case. I gave the planet a huge budget like 99 or so, so there shouldn't be a shortage or money for the biospheres.

Other planets, with worse population per biosphere and lower richness, don't delete their biospheres. I don't know what makes this planet different.

It would be ideal if you could just tell the colony blueprint to fill the planet with biospheres or not and how many tanks/troops the governor should build.

Ludoal:
Fixes #313.

The scrap decision for free biospheres never consulted the colony
budget: TryBuildBiospheres gates CONSTRUCTION on budget, but the
demolition branch keyed on profitability alone, so governors razed
player-built biospheres even at budget 99. If the budget covers the
upkeep of the existing free biospheres, that is an explicit 'I am
paying, keep them' - no scrapping. Tight-budget governors keep their
current cleanup behavior.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet_EvaluateBuildings.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet_EvaluateBuildings.cs
@@ -821,9 +821,14 @@ namespace Ship_Game
                 if (NumFreeBiospheres > 0)
                 {
                     // We do not need more than 1 free biospheres if not profitable.
-                    // We need only 1 free biosphere if we have anything to built at all
-                    shouldScrapBioSpheres = NumFreeBiospheres > 1 
-                        || numBuildingsWeCanBuild == 0 && (!HasBlueprints || Blueprints.IsAchievableCompleted);
+                    // We need only 1 free biosphere if we have anything to built at all.
+                    // But a colony budget that covers the upkeep of the free biospheres is an
+                    // explicit 'I am paying, keep them' - the scrap decision never consulted
+                    // the budget, so governors razed player-funded biospheres at budget 99 (issue 313)
+                    bool budgetCoversUpkeep = budget >= NumFreeBiospheres * bio.ActualMaintenance(this);
+                    shouldScrapBioSpheres = !budgetCoversUpkeep
+                        && (NumFreeBiospheres > 1 
+                            || numBuildingsWeCanBuild == 0 && (!HasBlueprints || Blueprints.IsAchievableCompleted));
                     return false;
                 }
                 else if (numBuildingsWeCanBuild == 0 || HabiableBuiltCoverage.Less(1))


gkapulis:
Your rule has landed as part of **91c6f0a1e** on `jupiter-1.60/fixes_33`, and issue #313 is closed with it. Closing this PR rather than merging, because the function it patches no longer exists: `933196b4f` replaced the tax-rate biosphere decision with a capacity model (issue #321, see the notes on #397), which deleted `BioSphereProfitable`, `numBuildingsWeCanBuild` and `HabiableBuiltCoverage` — every line this diff touches.

The principle you identified was right and it is what fixes #313: the scrap side never consulted the budget at all, while the build side did. Worth recording two things that came out of carrying it over, because neither was visible from the diff alone.

**The batch is what triggers it.** Hand-ordering several biospheres *raises* the colony's population cap, which *drops* the population ratio — the population has not grown into the new room yet, which is the whole reason for building them. Under the new rules that lands squarely in the excess-capacity branch, so the budget check has to come before the "more than one free biosphere" branch, not after it. Ordering is the fix, not just the presence of the test.

**The comparison is against zero, not against the upkeep.** This one caught me too — I wrote `budget >= NumFreeBiospheres * bioUpkeep` first, the same shape as yours, and it was wrong. The budget handed to the governor is `PlanetBudget.RemainingCivilian`, computed in `Budget.cs:60` as `CivilianAlloc - P.CivilianBuildingsMaintenance`, and `CivilianBuildingsMaintenance` is `Money.Maintenance - GroundDefMaintenance` — which already includes the standing biospheres. So comparing it against their upkeep again asks for enough headroom to fund a *second* identical set. With the reporter's manual 99 it happens to pass, so #313 as filed would have looked fixed; on a colony with a normal allocation `RemainingCivilian` sits near zero and the branch would never have fired, leaving the batch trimmed one per turn. `budget >= 0` is what "the colony is paying for them" actually means once the upkeep is already inside the number.

A second problem turned up while tracing the same report. Scrapping a biosphere could spend a **player-built** one while governor-built ones were standing. A biosphere has no tile of its own — it sets `tile.Biosphere` and puts a `Building` into the planet-wide list with nothing tying the two together — so `ClearBioSpheresFromList` removed whichever `FindBuilding` returned first. `TryScrapBiospheres` only guards planet-wide (is there *any* governor-built biosphere here), which is not the same question. It now prefers a governor-built one, the nearest thing to `PlayerBuiltIsProtected` that a pool without per-tile ownership allows.

Both are covered by tests, each verified by re-introducing the bug and watching the matching test fail.

Thank you for this one — the diagnosis was sound and it survived a rewrite of everything around it.

gkapulis:
Rule landed in 91c6f0a1e; the function this patched was replaced by 933196b4f. See the previous comment.

PR #370, opened 2026-07-22
Biosphere profitability judged at a nominal tax rate

### LINKED ISSUE #321: The construction of biospheres to increase the population should be independent of the tax rate

Ludoal:
Fixes #321.

NetRevenueGain multiplied by the live empire tax rate, and the governor
uses it on BOTH sides: build when profitable, scrap when not (via the
issue-313 path). High tax -> build, low tax -> raze: a structural
build/scrap oscillation driven by fiscal policy (worse with automatic
tax balancing, and it hurts the AI too). Infrastructure decisions now
use a fixed nominal 25% rate, decoupled from the tax slider.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/ColonyResource.cs
+++ b/Ship_Game/Universe/SolarBodies/ColonyResource.cs
@@ -299,7 +299,12 @@ namespace Ship_Game.Universe.SolarBodies
             if (b.IsBiospheres)
                 newPopulation += Planet.PopPerBiosphere(Planet.Owner)*0.001f;
 
-            float grossIncome = newPopulation * IncomePerColonist * TaxRate;
+            // Judge infrastructure at a fixed NOMINAL tax rate: keying on the live empire
+            // tax rate made governors build biospheres at high tax and raze them at low
+            // tax - a structural build/scrap oscillation driven by fiscal policy, since
+            // the scrap side (BioSphereProfitable) uses this very value (issue 321)
+            const float NominalTaxRate = 0.25f;
+            float grossIncome = newPopulation * IncomePerColonist * NominalTaxRate;
             return grossIncome - b.ActualMaintenance(Planet);
         }
 


gkapulis:
Closing this one, but the diagnosis is a good catch and we're taking it further rather than dropping it.

**You're right about the mechanism.** `NetRevenueGain` feeds both sides of the decision — the build side via `CanAffordManualControlled` (Planet_WeCanAffordThis.cs:122) and the scrap side via `BioSphereProfitable` -> `shouldScrapBioSpheres` (Planet_EvaluateBuildings.cs:818). Because both read the same value, moving the tax rate moves both thresholds together: build at high tax, raze at low tax. With automatic tax balancing nudging the rate every turn, that is a genuine structural oscillation, and it hits the AI hardest since its rate moves constantly. Good find.

**Two reasons we're not taking the fixed-constant version:**

1. **It drops more than fiscal policy.** `ColonyMoney.TaxRate` is not just the tax slider — `Update()` sets it from `data.TaxRate` at line 309 and then multiplies it at line 327 by `taxRateMultiplier`, which accumulates the racial `Traits.TaxMod` (line 314) and every `PlusTaxPercentage` building on that specific planet (line 319). Replacing the whole field with a bare `0.25f` throws those away too: a merchant race's tax trait stops making biospheres more attractive, and tax buildings on a planet stop influencing that planet's own biosphere decisions. Those are permanent, non-oscillating factors that should stay in the calculation.

2. **A fixed anchor is wrong for most of the game.** 0.25 is well chosen in one sense — it is exactly the default starting rate (EmpireData.cs:419) — but empires routinely run well above it in the mid game, so judging at a frozen 25% makes biospheres look systematically less profitable than they are, building fewer and scrapping more.

**What we'll do instead:** smooth the tax rate rather than freeze it. This codebase already leans on `HelperFunctions.ExponentialMovingAverage` (HelperFunctions.cs:313) for exactly this class of problem — `AverageFreighterCargoCap`, `AveragePlanetStorage`, `AverageFoodImportTurns`, `Relationship.Threat`, `MilitaryScore`, and every AI budget allocation in `RunEconomicPlanner` are all `[StarData]` fields nudged once per turn. Adding an `AverageTaxRate` alongside them, and having the biosphere check use that smoothed rate multiplied by the planet's real `TaxRateMultiplier`, kills the oscillation without discarding racial and building bonuses — and removes the anchor question entirely, since the empire's own sustained rate becomes the anchor.

We'll implement that on our side in the near future. Thanks for isolating the cause — the fix would not have been obvious without it.

Note that #371 is unaffected and being handled separately; it stands on its own regardless of what drives the biosphere flag.

PR #371, opened 2026-07-22
Close the colonist import/export overlap window

### LINKED ISSUE #293: Trading population back and forth between planets
Image

I think there are mixed things happening here. I don't know the exact logic, but it is something like:

If above 90% pop and above 2 billion pop, then see whether you can export population.
If below 80% pop, then see whether you can import population.
If above 95% pop and building a biosphere, then assume that we are expanding the population cap and start importing population with a single freighter.

1. Ping ponging population situation 1 where rule 1 and 3 conflict:

The first rule and third rule happened here just after each other. The first one made the planet export population to less populated planets. The third one kicked in when a biosphere was built to make room for the recently invented xeno mine. So now the planet is ping-ponging between importing and exporting population.

I don't quite see the use for rule 3. If you are massively building biospheres and they offer a lot of room, then you will drop below 80% of the population and start importing population. If the biospheres only offer a small amount of population or you're not building them very fast, then the population growth may easily keep up. The single imported population transport while building biospheres will often be far too little on industrial planets building massive biospheres and too much on research planets building small biospheres. So rule 1 and 2 are enough.

2. Ping ponging population situation 2, where rule 1 and 2 conflict:

Also, to avoid ping-ponging population in and out, you should not export population below 80% of the population. So when a 3 billion population planet goes above 2.7 billion people (90% of 3 billion), then it starts exporting people. But when it drops below 80%, which is 2.4 billion people, then it starts importing again. The freighter that visits the planet and finds it at 2.73 billion people should not export more than 2.73-2.4 billion =330 million people even if it is a large freighter that can carry 400 million people. These larger numbers of people per freighter are more prevalent when a race has the manifest destiny trait. I haven't played with that trait in a while, but have seen this issue in the past.

3. Importing very few people to huge worlds:

In the mid game, I regularly colonise these planets that were liberated from the Remnant while I have many dozens of planets on export population. You sometimes acquire these planets with a population cap of 20 billion and then they get a maximum of 5 transports at a time of say 50 million people. In the end, to fill it, you'll need 400 transports of 50 million people to get to that 20 billion, so that will take a while. I don't see why the import should be limited like that. You don't want the people transportation to drown out all other forms of transport of goods, but if there are superfluous freighters, they should bring people to this huge world. I sometimes manually increase the number of freighters assigned to this job in these huge planets, but that is a lot of micromanagement and you need to shut it down again at some moment.

Ludoal:
Fixes #293.

ShouldImportColonists allowed biosphere-anticipation imports up to a
0.99 population ratio while the export rule starts at 0.9: in the
]0.9, 0.99[ window on a large planet, the trade direction depended
solely on BiosphereInTheWorks, so every biosphere queued flipped the
traffic and freighters already in flight kept coming - simultaneous
import and export, the reported ping-pong. Bounding the anticipation
rule at the export threshold keeps its real purpose (fill capacity
under construction while below export levels) and removes the window.
Best paired with the issue-321 fix, which stops the biosphere
build/scrap oscillation that kept flipping the flag.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet_Resources.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet_Resources.cs
@@ -85,7 +85,10 @@ namespace Ship_Game
             return GoodState.STORE;
         }
 
-        bool ShouldImportColonists(float popRatio) => popRatio < 0.8f || BiosphereInTheWorks && PopPerBiosphere(Owner) > 100 && popRatio < 0.99f;
+        // the biosphere-anticipation import rule is bounded at the EXPORT threshold (0.9):
+        // in the old ]0.9, 0.99[ window import and export were both defensible and every
+        // biosphere queued flipped the traffic direction - the population ping-pong (issue 293)
+        bool ShouldImportColonists(float popRatio) => popRatio < 0.8f || BiosphereInTheWorks && PopPerBiosphere(Owner) > 100 && popRatio < 0.9f;
 
         public bool ShortOnFood()
         {


gkapulis (approved):
Verified: ColonistsTradeState checks import before export, so with BiosphereInTheWorks true and a ratio in ]0.9, 0.99[ line 83 returns IMPORT and line 84 is never reached - the moment the flag clears, the same planet returns EXPORT. And BiosphereInTheWorks is recomputed every governor pass from BuildingInQueue(BiospheresId) (Planet_Govern.cs:115), so anything queuing or dequeuing a biosphere flips the traffic direction. Bounding at 0.9 gives a clean boundary: at exactly 0.9 neither rule fires and the planet stores, so no overlap and no new gap. Single caller, and the export rule is gated on MaxPopulation > 2000, so this only ever affected large planets as you describe.

One precision note: a planet can't literally import and export at once, since the function returns a single state - what you observe is the state flipping while freighters dispatched under the previous state are still arriving. Same symptom, but worth being exact about.

Merging, comment trimmed on the branch.

**We are deliberately NOT closing #293.** The reporter raises three distinct problems and this fixes the first: (2) export above 90% vs import below 80%, where one large freighter can carry more than the delta and drop the planet straight into import range - he specifically calls out Manifest Destiny making this worse - and (3) huge worlds importing via only ~5 transports of 50M, needing hundreds of trips. Both remain open. I've noted that on the issue.

Also worth knowing: PR #370 is closed. Your diagnosis there was right - NetRevenueGain drives both the build and scrap sides so tax movement toggles biospheres - but a fixed constant discards the racial TaxMod and per-planet PlusTaxPercentage bonuses that ColonyMoney.TaxRate folds in, so we're implementing an exponential moving average of the tax rate instead, matching the idiom already used for budgets and threat. Details in that PR.

PR #372, opened 2026-07-22
Terraforming status clears when an event terraformer dies

### LINKED ISSUE #294: Terraforming status of the World tree remains after it has run out of its resource
Image

As you can see in this picture, the planet is still in the terraforming status. But the World Tree that was terraforming has gone when its food resource cache ran out.

Ludoal:
Fixes #294.

UpdatePlanetStatsFromRemovedBuilding zeroed TerraformPoints only for
regular terraformers (b.IsTerraformer); a depleted World Tree is an
event terraformer, so its removal left the points frozen and the planet
displayed 'Terraforming in Progress' forever (ApplyTerraforming's
early-outs never got to clean it once the planet had nothing normally
terraformable left). Same one-line rule as the TerraformingHere line
just above it.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet_ConstructionQueue.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet_ConstructionQueue.cs
@@ -457,7 +457,9 @@ public partial class Planet
             TerraformingHere = HasBuilding(bb => bb.IsTerraformer || bb.IsEventTerraformer);
         
         // FB - no terraformers present, terraform effort halted
-        if (b.IsTerraformer && !TerraformingHere)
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified: the TerraformingHere recompute one line above already counts both terraformer kinds - only the points-zeroing missed IsEventTerraformer, and ApplyTerraforming's !Terraformable early-out returns without cleanup, so the frozen points had no self-heal path (the colony screen shows 'Terraforming in Progress' purely on TerraformPoints > 0). Guard still requires !TerraformingHere so mixed-terraformer cases keep their progress. Merging - will trim the comment in a follow-up on the branch.

PR #373, opened 2026-07-22
DSB menu allows platforms in orbit of planets close to the sun

### LINKED ISSUE #284: Platforms cannot be placed in the orbit of planets that are close to a star using the deep space building menu.
Issue: Platforms cannot be placed in the orbit of planets that are close to a star using the deep space building menu, but they do get placed there when you order them through the planet.

Ludoal:
Fixes #284.

The deep space build menu rejected any spot within 20k of the star, so
planets orbiting inside that radius could never receive platforms from
this menu while the planet UI path allowed them. The sun-distance guard
now yields when the cursor is snapped to a planet; the radiation-safety
guard is unchanged.

### DIFF
--- a/Ship_Game/DeepSpaceBuildingWindow.cs
+++ b/Ship_Game/DeepSpaceBuildingWindow.cs
@@ -185,6 +185,8 @@ namespace Ship_Game
                 return false;
 
             if (targetSystem != null && (Screen.CursorWorldPosition2D.InRadius(targetSystem.Position, MinimumBuildDistanceFromSun) 
+                                         && targetPlanet == null // a planet's orbit is a valid site even close to the sun -
+                                                                 // the planet UI allows it, this menu refused it (issue 284)
                                          || !targetSystem.InSafeDistanceFromRadiation(Screen.CursorWorldPosition2D)))
             {
                 return false;


gkapulis:
Landed as **b0769a00b** on `jupiter-1.60/fixes_33`. Closing here since the branch carries it.

The premise checks out, and it is reachable on ordinary maps rather than exotic ones: the first planet ring sits at thirty times the star radius, so 7500 to 15000 units out, comfortably inside a 20000 standoff.

Worth recording what that guard actually was, since it looks load-bearing and is not. **I added it myself**, in `b82d4e9db`, in the same hunk that changed the red ring to `SunDangerRadius.LowerBound(MinimumBuildDistanceFromSun)` — it is a floor so a non-radiating star does not draw a ring of zero radius. It was never a "do not build inside the star" rule; the sun sprite spans 40000 to 150000, so the inner planets are already drawn inside it, and radiation is the separate check you correctly left alone.

The part I had not appreciated until reading the call sites is that this was not merely a refusal, it was a refusal that contradicted itself. The planet-ring and tether-line checks pass no system, so the sun clause never applied to them — the menu drew the orange buildable ring and printed "Will Orbit &lt;planet&gt;" while the ghost under the cursor stayed red and the click played NegativeClick. Your two lines make those agree.

Two changes from your version, both small:

- The comment came out. Reasoning goes in the commit message in this repo.
- `targetPlanet == null` moved to the front of the clause. Semantically identical — `InRadius` is `[Pure]` and `CursorWorldPosition2D` is a pure unproject — it just short-circuits one unproject and reads as a guard.

The Codex entry needed updating alongside, which is the part a reviewer is for: `CodexExpansionDeepSpaceBuildingText` stated the standoff and the radiation zone as one sentence. They are now separate, because after this only one of them depends on being near a planet.

**One thing I deliberately did not take**, in case you want it: the red danger ring at `DeepSpaceBuildingWindow.cs:271` is still drawn per system at `SunDangerRadius.LowerBound(20000)`, so the ghost now turns green inside a ring that says no. I left it because the ring is not cursor-dependent and making it so would flicker as the cursor crossed planets, and because "no free-space building here" is still true of it. If you disagree, that is a reasonable follow-up.

**And the half of the asymmetry this does not close**, which your PR led me to: the colony-screen path has no radiation check either. `Planet.AddOrbital` creates the goal with no distance and no radiation test, while `ApplySolarRadiationDamage` damages every non-Guardian ship in the system's `ShipList`, orbitals included. So a colony inside a pulsar's or a black hole's zone keeps building orbitals that are then cooked. It is worse than the planet's own orbit suggests, because `FindNewOrbitalLocation` picks a random angle up to about 6000 units out and `TetherOffset` is fixed in world space, so the distance to the star swings as the planet orbits. That is logged as its own task rather than folded in here — it is a gameplay change with a real blast radius, and it deserves a test.

Thanks for this one. A rule that two screens disagreed about is the kind that players report as "sometimes it works".

PR #374, opened 2026-07-22
Clarify money units: all flows are per turn

Ludoal:
- Espionage screen: cost label bc/y -> bc/turn (backing field is CostPerTurn)
- Governor civilian budget tooltip: bc/y -> bc/turn (ENG + UKR)
- Economic Overview: subtitle noting all money values are per turn
- Treasury Goal tooltip: clarify 20 year projection = 200 turns

Money is charged every turn while some labels implied per star-year
(10 turns). Maintainer feedback.

Test status: labels only; verified in-game on our fork.

### DIFF
--- a/Ship_Game/Data/GameText.cs
+++ b/Ship_Game/Data/GameText.cs
@@ -4096,17 +4096,17 @@ namespace Ship_Game
         Override = 4226,
         /// <summary>Override this budget and set your own.</summary>
         OverrideThisBudgetAndSet = 4227,
-        /// <summary>Civilian Buildings Expenditure/Budget in BY/C. If you see over-budget, you</summary>
+        /// <summary>Civilian Buildings Expenditure/Budget in bc/turn. If you see over-budget, you</summary>
         CivilianBuildingsExpenditurebudgetInByc = 4228,
-        /// <summary>Ground Defense Buildings Expenditure/Budget in BY/C. If you see over-budget,</summary>
+        /// <summary>Ground Defense Buildings Expenditure/Budget in bc/turn. If you see over-budget,</summary>
         GroundDefenseBuildingsExpenditurebudgetIn = 4229,
-        /// <summary>Orbitals Expenditure/Budget in BY/C</summary>
+        /// <summary>Orbitals Expenditure/Budget in bc/turn</summary>
         OrbitalsExpenditurebudgetInByc = 4230,
-        /// <summary>Civilian Buildings Expenditure in BY/C</summary>
+        /// <summary>Civilian Buildings Expenditure in bc/turn</summary>
         CivilianBuildingsExpenditureInByc = 4231,
-        /// <summary>Ground Defense Buildings Expenditure in BY/C</summary>
+        /// <summary>Ground Defense Buildings Expenditure in bc/turn</summary>
         GroundDefenseBuildingsExpenditureIn = 4232,
-        /// <summary>Orbitals Expenditure in BY/C</summary>
+        /// <summary>Orbitals Expenditure in bc/turn</summary>
         OrbitalsExpenditureInByc = 4233,
         /// <summary>Total:</summary>
         Total3 = 4234,
--- a/Ship_Game/Data/GameText.cs
+++ b/Ship_Game/Data/GameText.cs
@@ -4096,17 +4096,17 @@ namespace Ship_Game
         Override = 4226,
         /// <summary>Override this budget and set your own.</summary>
         OverrideThisBudgetAndSet = 4227,
-        /// <summary>Civilian Buildings Expenditure/Budget in bc/turn. If you see over-budget, you</summary>
+        /// <summary>Civilian Buildings Expenditure/Budget in bc/t. If you see over-budget, you</summary>
         CivilianBuildingsExpenditurebudgetInByc = 4228,
-        /// <summary>Ground Defense Buildings Expenditure/Budget in bc/turn. If you see over-budget,</summary>
+        /// <summary>Ground Defense Buildings Expenditure/Budget in bc/t. If you see over-budget,</summary>
         GroundDefenseBuildingsExpenditurebudgetIn = 4229,
-        /// <summary>Orbitals Expenditure/Budget in bc/turn</summary>
+        /// <summary>Orbitals Expenditure/Budget in bc/t</summary>
         OrbitalsExpenditurebudgetInByc = 4230,
-        /// <summary>Civilian Buildings Expenditure in bc/turn</summary>
+        /// <summary>Civilian Buildings Expenditure in bc/t</summary>
         CivilianBuildingsExpenditureInByc = 4231,
-        /// <summary>Ground Defense Buildings Expenditure in bc/turn</summary>
+        /// <summary>Ground Defense Buildings Expenditure in bc/t</summary>
         GroundDefenseBuildingsExpenditureIn = 4232,
-        /// <summary>Orbitals Expenditure in bc/turn</summary>
+        /// <summary>Orbitals Expenditure in bc/t</summary>
         OrbitalsExpenditureInByc = 4233,
         /// <summary>Total:</summary>
         Total3 = 4234,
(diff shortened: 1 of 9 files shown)


gkapulis (changes requested):
Please change bc/turn to bc/t  and BC/T

Ludoal:
Done - bc/turn reads bc/t everywhere now (34 sites: yaml, the six screens, and the enum comments kept in sync). Thanks for the review.

Ludoal:
All 34 call sites updated back in August. Ready for your eyes whenever you have a moment.

gkapulis (changes requested):
Thanks for the round 2. The C# side is sound and the `BudgetPercent` repositioning is correct — `FontBig` really is the font `BudgetSum` is created with, so that width math holds. Four things before this can go in.

## 1. Casing — please preserve it per site

My earlier note asked for `bc/t` **and** `BC/T`, meaning keep whatever casing each site already had. Everything got lowercased instead. In the base branch 11 of the 12 display sites are uppercase and one is lowercase:

**These were `BC/Y`, so they should read `BC/T`:**
- `DeepSpaceBuildingWindow.cs` — maintenance cost
- `ColonyBlueprints/BlueprintsBuildableListItem.cs` — `maintString`
- `ColonyScreen/BuildableListItem.cs` — `maintString`
- `ColonyScreen/ColonyScreen_Draw.cs` — gross income, gross upkeep, net income (3)
- `ColonyScreen/GovernorDetailsComponent.cs` — `BudgetSum` in both branches, plus the three `NoGovernor*Expense` labels (5)

**This one was `bc/y`, so `bc/t` is right as-is:**
- `Espionage/EmpiresButton.cs` — `CostPerTurn`

The yaml is mixed in the same way and predates your PR: 4228's ENG was lowercase `bc/y`, while 4229-4233 were uppercase `BY/C`. Same rule there — preserve each as it was. If you think the mixed casing is itself worth unifying, that is a fair argument, but make it a separate commit so it is reviewable on its own.

## 2. Russian is untouched, and it is a shipped language

`Language.Russian` is in the enum and `Localizer.cs:159` reads the `RUS` column, so this is live for players. All six budget tokens still carry per-year units:

```
4228 RUS  ... в МГ/K (Миллиард Кредитов/Год)
4229 RUS  ... в МК/Г (Миллиард Кредитов/Год)
4230 RUS  ... в МК/Г
4231 RUS  ... в МК/Г
4232 RUS  ... в МК/Г
4233 RUS  ... в МК/Г
```

Two of them spell it out — *Billion Credits/**Year***. A Russian player reads exactly the bug this PR exists to remove, and now they are the only ones who do, which is worse than before the change was uniform.

You kept the Latin `bc/t` in the Ukrainian and Portuguese strings rather than inventing a localized abbreviation. That is a good call, so do the same for Russian: `МК/Г` and `МГ/K (Миллиард Кредитов/Год)` both become `bc/t`. Mechanical, and consistent with the other two columns.

(Note 4228 says `МГ/K` while the rest say `МК/Г` — the letters are transposed. Pre-existing, and it disappears with this replacement anyway.)

## 3. Two translations the sweep could not catch, and they are the point

`4231 UKR` and `4232 UKR` were missed, and they are worth looking at:

```
4231 UKR: "Видатки на цивільні будівлі в білоруських рублях"        -> "in Belarusian rubles"
4232 UKR: "Витрати на наземні оборонні споруди в британських доларах" -> "in British dollars"
```

A translator read `BY/C` as the Belarus ISO currency code, and `BY` as "British". These are the strongest argument your PR has: the ambiguous unit did not merely confuse players, it corrupted the translation into a different currency entirely. They are also exactly what a search for `BY/C` cannot find, because the token is already gone — which is the general lesson here. When a sweep is driven by matching the old string, the strings that were mangled worst are the ones that will not match.

Both should read `bc/t` like their siblings.

Also `4228 PTB` still reads `em bc/ano` while 4229-4233 PTB were all fixed — same tooltip family, so it stands out.

And `TreasuryGoalIsTheTarget` (1918) got the "(200 turns)" clarification in ENG only; RUS, UKR and PTB still say a bare 20-year projection. Same fix belongs in all four.

## 4. The new Economic Overview label is hardcoded English

```cs
string unitNote = "(all money values are per turn)";
```

This sits three lines below a title that goes through `Localizer.Token(GameText.EconomicOverview)`. It compiles only because `LocalizedText` takes an implicit conversion from `string`, so nothing warns you. Every non-English player gets that line in English — in a PR whose whole purpose is making units legible in every language.

It needs a real `GameText` token with all shipped columns filled. Since it is new text rather than a reworded existing one, it also needs a new id rather than reusing one.

## Nits

- `BudgetScreen.cs` drops a blank line after the title `Label(...)` — unrelated to the change.
- The new `Label(...)` is wedged between `EmpireNetIncome.DropShadow = true;` and `EmpireNetIncome.DynamicText = ...`, splitting that block. Move it after the `DynamicText` assignment.
- `BudgetPercent.Pos` is now set both at layout time (`GovernorDetailsComponent`, the `TopLeft.X + CivBudgetRect.Width + 15` line) and again in `UpdateBudgets`. The layout one is dead whenever a governor is on. Not wrong, but worth a glance so the two do not drift.

## Confirmed fine, leave these alone

The `GameText.cs` doc comments match what the generator would emit, and the enum names are safe — `LocalizationDB.LoadIdentifiers` matches existing tokens by `Id`, so the stale `...InByc` names will not be renamed out from under us on a regeneration. No need to touch them.

Once the four above are done this is good to go — the remaining work is all mechanical.

gkapulis:
Landed as **51a12578d** on `jupiter-1.60/fixes_33`. Closing this and finishing it in house rather than sending you round three — you had already done two rounds on it, and what was left was mostly translation archaeology that is easier for me to do against the file than to describe in a review comment.

**The diagnosis is yours and it was right.** Upkeep, income and the governor budgets are all collected once a turn, and twelve labels said `BC/Y`. A star year is ten turns, so every one of them implied a rate ten times off. That is now `BC/T`.

Two things went differently from your round 2.

**Casing is preserved per site.** Round 2 lowercased all twelve to `bc/t`; eleven of them were upper case and only the espionage cost in `EmpiresButton.cs` was not. That also retired the `BudgetPercent.Pos` line you added — it was only needed because lowercase is narrower, and `BC/T` is the same width as `BC/Y`, so the existing fixed column still lines up.

**All four languages, not just ENG and UKR.** This is the part that needed digging, and the six tooltips turned out to be worse than stale — they disagreed with each other. ENG had both `bc/y` and `BY/C`. PTB had `bc/ano`. RUS spelled out `Миллиард Кредитов/Год` under two different abbreviations, `МГ/K` and `МК/Г`.

They all name `BC/T` now, in every language, which is a change of approach worth explaining: **the unit on screen is a hardcoded literal that is never localized.** A Russian player reads `BC/T` on the colony screen, so a tooltip explaining `МК/Г` was describing a unit that does not appear anywhere in the game. The spelled-out Russian expansion stays, updated to `за ход`.

**Two of the Ukrainian values were not translations at all.** Id 4231 said civilian building spending was measured «в білоруських рублях» — in Belarusian rubles — and 4232 said ground defense was «в британських доларах», in British dollars. Somewhere upstream a translator read `BY/C` as a currency code. I mention it because it explains why your pass could not have caught them: no search for the old unit string will ever match a value that no longer contains it. If you ever touch localization here again, that is the failure mode to watch for.

**The Economic Overview note.** You were right that the screen needs one — its figures go through `MoneyString()` and carry no unit at all, so it is the one place a player cannot infer the rate. It is a tooltip on the screen title rather than the subtitle, though. The subtitle sat at `Y+38`; `taxRect` starts at `Y+37`, and `SummaryPanel` builds with no header and `Padding.Y` of 2, so the tax slider draws its own caption from `Y+39` and the two would have shared a row — worse in RUS, UKR and PTB, where the slider caption is longer but the note stays English. A visible line wants a genuinely free strip and someone looking at the screen; the title tooltip costs nothing meanwhile. New token 4576, hooked to the budget Codex entry so F1 reaches it.

One piece of your PR was already in, and better: the treasury slider tooltip was rewritten during the Codex pass and now says "this fraction of a two hundred turn projection", which beats "20 year (200 turns)".

Left alone on purpose: the enum names still end `InByc`. They bind by Id, so renaming would drag the yaml NameIds and `CodexHooks.yaml` along for no behavioural gain.

One thing your PR raised that I have not acted on, and that is a design call rather than a bug: `MoneyString()` is a plain two-decimal format with no scaling anywhere, and the Codex economy entry says "credits per turn". So the `B` in `BC` may be as wrong as the `Y` was. I did not want to redenominate the currency inside a units fix.

Thanks for pushing on this twice. Labels that quietly state the wrong rate are the sort of thing players learn to distrust rather than report.

PR #375, opened 2026-07-22
Research screen: anchor the Search button to the frame bottom

Ludoal:
`searchPos` uses `main.Height` as a Y coordinate — harmless today because the frame starts at zero, but it is `main.Bottom` that was meant, and any future offset of the research frame silently misplaces the button (we hit it on our fork).

Test status: no visible change upstream by construction; verified in-game on our fork with an offset frame.


gkapulis (approved):
Verified: main is Rectangle(0, 0, ScreenWidth, ScreenHeight) in LoadContent, so Height == Bottom today and this is a no-op upstream by construction — pure correctness hardening. The X term on the same line already uses the main.X + main.Width idiom, which makes the missing main.Y on the Y term a clear oversight. Zero risk.

PR #376, opened 2026-07-23
Governor no-scrap setting protects player buildings from Replace

### LINKED ISSUE #303: Planetary governor setting 'governor will not scrap buildings' does not work
[Mars__1.51.15106_Combined_Arms-v8.7i_The Draylok Council, Star Date 1028.1.zip](https://github.com/user-attachments/files/28935130/Mars__1.51.15106_Combined_Arms-v8.7i_The.Draylok.Council.Star.Date.1028.1.zip)

The planetary governor setting 'governor will not scrap buildings' does not work. It will still remove your manually placed buildings if it finds another building more useful. So, it doesn't match with its explanation text.

I think it has some overlap with the 'specialised trade hub' setting. But the 'governor will not scrap buildings' should still allow the governor to add buildings, just not remove them. While the 'specialised trade hub' leaves all building construction to the player. But as said, that's not how it works. The 'governor will not scrap buildings' setting still allows the governor to scrap buildings.

In the attached savegame, the planet of Drell VI is building a research lab on the one tile that is available to it. It has the option enabled that the 'governor will not scrap buildings'. Once it is completed the manually selected building, it immediately scraps it and replaces it with another.

Ludoal:
Fixes #303.

Two stacked holes made the setting a no-op exactly where the reporter hit it:
- `ReplaceBuilding` never consulted the per-design `DontScrapBuildings` setting (`TryScrapBuilding` and the over-budget path do). A replace is a scrap with a gift behind it, so the setting now covers it.
- The *unless they are player built* guard in `SuitableForScrap` sat after the no-blueprint early return — without blueprints every building exited as scrappable one line earlier, so the player guard was dead code. It moves to the head of the method, unconditional.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet_EvaluateBuildings.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet_EvaluateBuildings.cs
@@ -339,7 +339,10 @@ namespace Ship_Game
             if (BuildingsHereCanBeBuiltAnywhere || BuildingsCanBuild.Count == 0)
                 return;
 
-            // Replace works even if the governor is not scrapping buildings, unless they are player built
+            // a replace is a scrap with a gift behind it: the no-scrap setting covers it too (upstream issue 303)
+            if (GovernorShouldNotScrapBuilding)
+                return;
+
             float worstBuildingScore = ChooseWorstBuilding(overBudget, scrapZeroMaintenance: true, true, out Building worstBuilding);
             if (worstBuilding == null)
                 return;
@@ -483,6 +486,7 @@ namespace Ship_Game
             if (b.IsBiospheres
                 || b.IsMilitary
                 || !b.Scrappable
+                || b.IsPlayerAdded && OwnerIsPlayer // player-built is never the governor's to scrap — the guard below sat after the no-blueprint early return and was unreachable (upstream issue 303)
                 || b.IsSpacePort && Owner.GetPlanets().Count == 1 // Dont scrap our last spaceport
                 || b.BuildOnlyOnce
                 || b.PlusTerraformPoints > 0) // using this instead of IsTerraformer since some event building might also terraform without the terraformer building ID


gkapulis:
Merging this — the `SuitableForScrap` half is a good find and it is the real fix. I am adjusting the `ReplaceBuilding` half in-house rather than asking you for another round, since it turned on a design decision rather than on the code.

**What you found, and why it is bigger than the issue suggests.** The setting's own tooltip ends with:

> "Note that Governors will never scrap a building built personally by you."

That promise is unconditional — it does not depend on the checkbox at all. And the guard implementing it has been unreachable on most planets, exactly as you diagnosed: `RequiredInBlueprints(b)` is `Blueprints?.IsRequired(b) == true`, so with no blueprints it is false, `if (!RequiredInBlueprints(b)) return true;` short-circuits as scrappable, and the `b.IsPlayerAdded && OwnerIsPlayer` check one block below never runs. Hoisting it into the unconditional block is right, and it fixes the reporter's case on its own.

**Why the `ReplaceBuilding` early return is coming back out.** The comment you replaced was describing deliberate behaviour:

```cs
// Replace works even if the governor is not scrapping buildings, unless they are player built
```

The maintainer's rule, confirmed: the checkbox means the governor will not *scrap* buildings but is still allowed to *upgrade* them; player-built buildings are never touched regardless of the checkbox. Your `SuitableForScrap` change delivers the second half completely — both `TryScrapBuilding` and `ReplaceBuilding` reach it through `ChooseWorstBuilding`, so one guard covers scrapping and upgrading alike.

That leaves the early return blocking only one thing: the governor replacing *its own* buildings with better ones on a no-scrap planet. That is the behaviour the checkbox was deliberately written to allow, and it bites hardest in the reporter's own scenario — Drell VI has a single free tile, so the governor would be frozen on whatever it built first, permanently.

Also removing the now-dead duplicate of the guard at the old site, which your change makes unreachable.

**What auditing this turned up**, since it is the useful part for anyone reading later. Checking both rules against every governor scrap path found three more gaps that are not yours to fix:

- `TryScrapMilitaryBuilding` consults the checkbox nowhere at all, and `SuitableForScrap` excludes `b.IsMilitary` outright — so that path is the only one that scraps military buildings, and "Governor Will Not Scrap Buildings" has never protected a single one.
- Its two branches disagree with each other: the no-blueprints branch filters `!b.IsPlayerAdded`, the blueprints branch does not.
- `TryCancelOverBudgetCivilianBuilding` cancels queued buildings without checking `qi.IsPlayerAdded`, while `SBProduction` does check it in two other queue paths.

Same shape as the bug you found: a rule enforced in one place and quietly missing in its sibling. Those are going in as a follow-up on the same branch.

Thanks — the unreachable-guard diagnosis was precise, and it is the kind of thing that only shows up if you actually trace the early returns rather than reading the conditions in order.

PR #377, opened 2026-07-23
Governor builds biospheres again on ex-volcano worlds

### LINKED ISSUE #312: Governor doesn't recognise terraformed tiles or tiles that became available after a volcanic eruption as available to build on.
[Mars__1.51.15106_Combined_Arms-v8.7i_The Draylok Council, Star Date 1159.8.zip](https://github.com/user-attachments/files/28934891/Mars__1.51.15106_Combined_Arms-v8.7i_The.Draylok.Council.Star.Date.1159.8.zip)

The governor doesn´t recognise the availability of building plots that once were volcanoes and other non-buildable tiles but have since been terraformed. It doesn't build biospheres on these tiles. If you manually place a biosphere on such a tile, then the governor will start building constructions on the biosphered tile. This is with the use of colony blueprints, but I don't know if it is in any way related to that. See the colony of Jerez I in the attached file. 

I don't have the colony blueprints for this game anymore. Base a blueprint on what is on the planet and extend it with one extra building and see whether it gets constructed. If the bug is still there, then it should not build the biosphere needed for the building.

[Jupiter__1.60.00043_Combined_Arms-v1.60.0005_The Kulrathi Shogunate, Star Date 1168.7.zip](https://github.com/user-attachments/files/29069357/Jupiter__1.60.00043_Combined_Arms-v1.60.0005_The.Kulrathi.Shogunate.Star.Date.1168.7.zip)

Image

Another instance of the same bug: the planet of Rafias VI has had a volcanic eruption. Some time after the eruption a new tile became available and on another the lava disappeared, but the governor following a blueprint is not finishing the blueprint. It doesn't recognise the new tiles.

Ludoal:
Fixes #312.

`GetPreferredTile`'s fallback `Find` accepted tiles carrying a building or a queued item; `Enqueue` then rejected that exact tile on every governor pass, while perfectly valid terraformable tiles (e.g. flagged back after an eruption, Volcano.cs restores the flag) stayed excluded from preference. Net effect: the governor stalled forever on those worlds, and manual placement worked because it validates the actual tile — the reporter's exact experience.

The preferred tile is now enqueue-valid or null; null falls through to the existing random tile assignment, which accepts terraformables.

Test status: compiled and logic-reviewed against the issue's repro (save Jerez I, Star Date 1159.8); play-testing on our fork pending.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet_EvaluateBuildings.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet_EvaluateBuildings.cs
@@ -851,9 +851,12 @@ namespace Ship_Game
                 PlanetGridSquare preferred = null;
                 if (Owner.IsBuildingUnlocked(Building.TerraformerId))
                 {
-                    preferred = TilesList.Find(t => !t.Habitable && !t.Terraformable && !t.BuildingOnTile);
-                    if (preferred == null)
-                        preferred = TilesList.Find(t => !t.Habitable && !t.Terraformable);
+                    // upstream issue 312: the old fallback accepted tiles carrying a building or a
+                    // queued item; Enqueue rejected them every pass while perfectly valid terraformable
+                    // tiles stayed excluded, so the governor never built the biosphere (manual placement
+                    // worked, it validates the actual tile). A null preferred falls through to random
+                    // tile assignment, which accepts terraformables.
+                    preferred = TilesList.Find(t => !t.Habitable && !t.Terraformable && !t.BuildingOnTile && t.NoQueuedBuildings);
                 }
                 else
                 {


gkapulis:
Applied in-house as **54e182a1f** on `jupiter-1.60/fixes_33`, with tests. Your diagnosis was exactly right and it is the part that took the longest to confirm: `AssignBuildingToTile` validates a tile it is handed and returns false rather than choosing another, so a preferred tile carrying a queued item fails the enqueue on every governor pass, forever. That is the ex-volcano stall, and the reason manual placement worked while the governor never got anywhere.

Two things made me rewrite rather than take the patch.

**The `else` branch has the same flaw.** With no terraformer unlocked the pick is `Find(t => !t.Habitable && !t.BuildingOnTile)` with a fallback to `Find(t => !t.Habitable)`, and that fallback checks neither the building nor the queued item — the identical bug, just on the other side of the branch. Fixing one side would have left a colony without terraformer tech stalling in the same way.

**The predicate can be exact instead of approximate.** Rather than adding `NoQueuedBuildings` to one `Find`, every candidate now has to satisfy `CanEnqueueBuildingHere(bio)`, which is literally the test `Enqueue` applies. That makes the whole class of bug unreachable instead of closing the one instance, and it collapses both branches into one method:

```csharp
internal PlanetGridSquare PreferredBiosphereTile(Building bio)
{
    bool saveGroundForTerraformer = Owner.CanTerraformPlanetTiles;
    return TilesList.Find(t => t.NoBuildingOnTile && t.CanEnqueueBuildingHere(bio)
                               && (!saveGroundForTerraformer || !t.Terraformable))
        ?? TilesList.Find(t => t.NoBuildingOnTile && t.CanEnqueueBuildingHere(bio))
        ?? TilesList.Find(t => t.CanEnqueueBuildingHere(bio));
}
```

Two details in there are worth calling out, because both cost me a round of review.

`t.NoBuildingOnTile` has to stay as a *preference*, and I had dropped it in my first attempt. `CanPlaceBuildingHere` for a biosphere is only `!Habitable && !VolcanoHere && !LavaHere` — it does not check for a building, so a biosphere may legally be placed on an uninhabitable tile that already carries a build-anywhere building (a terraformer, Mountain, Cave system, Crater, Gold Deposit). Roofing one of those buys the colony no room, and `FreeHabitableTiles == 0` is the main reason the method runs at all: the colony would have paid for a biosphere and still had nowhere to build. Bare ground first, occupied ground only as a last resort.

`Owner.CanTerraformPlanetTiles` replaces `Owner.IsBuildingUnlocked(TerraformerId)`. This is the same correction I asked for on #405: turning a tile is terraforming level 2, so between unlocking a terraformer and reaching level 2 there is nothing to save the terraformable ground for, and skipping it there only sends the biosphere further down the list for no reason.

Six tests now cover this pick and the terraformer's, four of them checked by reverting the change and confirming the matching test fails.

Thank you for this — it is a genuinely hard failure to spot from a save, and the reasoning about manual placement validating the actual tile is what pinned it down.

gkapulis:
Applied in-house as 54e182a1f with tests; see the previous comment for why the rewrite went wider than the patch.

PR #378, opened 2026-07-23
Scroll wheel moves content, not scrollbar pixels

Ludoal:
Wheel and scroll-button deltas were expressed in SCROLLBAR pixels (`EntryHeight/3` per notch). The bar maps its track to the whole list, so the real content jump scaled with list length and resolution: on a 1080p Empire screen with a long colony list, one wheel notch skipped rows entirely.

Deltas are content-space now — one entry per wheel notch, half an entry per scroll-button click — converted exactly through the track/content ratio. Affects every ScrollList consistently; scrollbar dragging is untouched.

Test status: verified in-game on our fork at 1080p (Empire screen, was skipping) and 1440p.

### DIFF
--- a/Ship_Game/UI/ScrollListBase.cs
+++ b/Ship_Game/UI/ScrollListBase.cs
@@ -133,8 +133,18 @@ namespace Ship_Game
 
         #region ScrollList HandleInput
 
-        int ScrollButtonAmount(bool held) => !held ? EntryHeight / 2 : EntryHeight / 6;
-        int ScrollWheelAmount => EntryHeight / 3;
+        // Ludoal fork: wheel/buttons used to move the SCROLLBAR by EntryHeight-derived pixels,
+        // so the real content jump scaled with list length and resolution (a notch skipped
+        // rows at 1080p on the Empire screen). Content-space deltas, converted exactly.
+        float ScrollBarDeltaForEntries(float entries)
+        {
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified: ScrollBarDeltaForEntries is the exact algebraic inverse of the bar->content mapping in PerformLayout (line 429-430), reusing the same maxVisibleItemsF expression, and it reduces to ScrollHousing.H * entries / Count so there is no blow-up on barely-scrollable lists. Dragging untouched, int->float widening safe (SetScrollBarPosition/ScrollBar.Y were already float), no external callers of the changed helpers.

One follow-up we'll push on the branch: ScrollWheelAmount can now return 0 for a list that fits its housing, so HandleInputMouseWheel returns false and the wheel is no longer captured - in UniverseScreen that falls through to HandleCameraZoomScrolling, i.e. wheeling over a short SolarsystemOverlay/ship panel would zoom the camera. Restoring capture-on-hover, plus the usual comment strip.

gkapulis:
Follow-up pushed as dc1b67e8e on jupiter-1.60/fixes_29: HandleInputMouseWheel now captures whenever a wheel notch lands on the list and scrolls only when there is somewhere to scroll, so a short list no longer leaks the event to the universe camera zoom. Comment stripped per our convention. Thanks for the fix - the content-space conversion itself is exactly right.

PR #379, opened 2026-07-23
Prioritized projectors queue below earlier projectors

### LINKED ISSUE #300: Subspace projectors should go to the top of the queue but below existing subspace projectors
The logic that places subspace projectors at the top of the queue also places it above subspace projectors which were just placed at the top of the queue. This way, you get one half subspace projector road and then another half road and maybe another half road before they get completed. 

They should be placed at the top of the queue, but below other subspace projectors. 

Priority build subspace projectors should go to the top of the queue, but below priority build subspace projectors which were already in the queue.

Ludoal:
Fixes #300.

Each new road node was moved to the absolute top of the queue, above
the projectors already prioritized — so a long road built as several
half-roads racing each other. New nodes now slot in after the road
nodes already leading the queue; rushed urgent orbitals keep the
absolute front.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/SBProduction.cs
+++ b/Ship_Game/Universe/SolarBodies/SBProduction.cs
@@ -492,11 +492,19 @@ namespace Ship_Game.Universe.SolarBodies
                     if (P.Owner.AutoBuildTerraformers && P.Owner.data.Traits.TerraformingLevel > 0 && !item.IsPlayerAdded)
                         DePrioritizeTerraformer();
 
-                    if (item.Rush && item.QType == QueueItemType.OrbitalUrgent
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified: the scan is bounded (new item sits at Count-1, loop indexes 0..Count-2 only), terminating, and runs under the ConstructionQueue lock; degenerate cases (non-road front -> index 0, all-roads -> self-position no-op) are safe. Road nodes now queue FIFO after the leading road run, which is the issue-300 fix. Note: a new road node still lands above a pre-existing rushed urgent orbital - same as the old absolute-front behavior, so no regression. Merging - will trim the comment in a follow-up on the branch.

PR #380, opened 2026-07-23
Crash sites wait for the conquest to resolve

### LINKED ISSUE #343: I am getting messages about the recovery of credits for ships on a planet that I am in the process of conquering
Image

I think that you should only recover ships on a planet that you own and not on a planet that you are still in the process of conquering. There is no infrastructure in place to recover anything on a planet that you do not own.

[Jupiter__1.60.00045_Combined_Arms-v1.60.0005_The Kulrathi Shogunate, Star Date 1210.3.zip](https://github.com/user-attachments/files/29561175/Jupiter__1.60.00045_Combined_Arms-v1.60.0005_The.Kulrathi.Shogunate.Star.Date.1210.3.zip)

This savegame is from a year before the screenshot. I can save at the moment of the screenshot but it is useless as messages are not part of the savegame. So you might be able to recreate the situation with the savegame of a year before I recover ships, but it might also be the case that the random number generator does not make ships crash on the planet that can be recovered.

Ludoal:
Fixes #343.

Landing troops on a crash-site tile of a planet someone else still
owns triggered the recovery mid-invasion (credits or even a spawned
ship for the invader). There is no recovery infrastructure on a world
you do not own yet: activation now waits until the planet is neutral
or yours. Deferring is safe — the trigger re-fires while a troop
holds the tile, so the site activates right after the conquest.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/PlanetGridSquare.cs
+++ b/Ship_Game/Universe/SolarBodies/PlanetGridSquare.cs
@@ -324,7 +324,10 @@ namespace Ship_Game
             {
                 if (IsCrashSiteActive)
                 {
-                    if (!planet.SpaceCombatNearPlanet)
+                    // upstream issue 343: no recovery infrastructure on a planet someone else
+                    // still owns — an invader waits for the conquest to resolve. Safe to defer:
+                    // this check re-fires as long as a troop holds the tile.
+                    if (!planet.SpaceCombatNearPlanet && (planet.Owner == null || planet.Owner == empire))
                         CrashSite.ActivateSite(planet.Universe, planet, empire, this);
                 }
                 else


gkapulis:
Landed as **68ca48ecc** on `jupiter-1.60/fixes_33`, with tests. Closing here.

The diagnosis is right and I confirmed it: `ActivateSite` takes `p.Owner ?? activatingEmpire`, and then either hands the activator the ship outright or, when the owner differs, pays them credits. So landing a troop on a crash site mid-invasion paid the invader on the defender's world. Roland's framing is the right one — there is no recovery infrastructure on a planet you do not own — and your gate is the correct shape for it.

**The gate on its own makes the situation worse, though, and that is why this went in with a second half.**

An event tile is a target: `HostilesTargetsOnTile` returns true for `EventOnTile` regardless of who owns the planet. `SpotClosestHostile` orders candidates by Manhattan distance, so the troop's *own* tile (distance zero) always wins, and `CalculateTargetValue` scores an event tile zero, so nothing else gets picked instead. `MoveTowardsTarget` then returns immediately because `ourTile == targetTile`.

Until now that resolved itself after one pass: the recovery fired, `DestroyBuildingOn` cleared the tile, and the troop moved on. With the recovery gated and nothing else changed, the troop that landed on the site stands there for the entire invasion — waiting for the conquest it has just stopped taking part in. So the same ownership condition is now on the targeting side too.

Blast radius on that: the defender's own troops and buildings scan with `spotterOwner == planetOwner`, so they are unaffected; unowned planets are unchanged; and an enemy troop standing on the site is still a target through the `LockOnEnemyTroop` fall-through. The tile stops being a magnet only when it is empty, which is what we want.

**Two consequences I dealt with rather than left behind:**

- The reward split keyed on that same `owner` variable, and once the gate is in place `owner` is provably always the activator (`ActivateSite` has exactly one caller). So the credits branch became unreachable — and worse, a recovery on an **unowned** planet would have pushed production into a store with no colony to spend it, quietly evaporating the reward. It now keys on the planet: your colony takes production, an unowned world pays cash.
- `DoExplorePlanet`'s validity check read `!Owner.IsAtWarWith(targetPlanet.Owner)`, and **`IsAtWarWith` returns false for yourself** — so it was already ending any exploration task aimed at our own planet, which is exactly where `DynamicCrashSite.SendTroop` falls back when no troop ship is free. My first attempt wrote the new rule as `Owner != Owner`, which would have quietly revived that dead path; `TaskStep` 5 calls `StartBombing` on whatever it is pointed at with no friendly-owner check. The condition is now simply "somebody owns it", which keeps our own planets ending as before and adds the new rule for everyone else's. Exploration tasks are only ever created for unowned planets anyway, so this is mostly a safety net.

A note on the test, since it is the kind of trap worth knowing about: `CheckAndTriggerEvent` only runs when one of *your* troops is on the tile (`LockOnOurTroop`). My first version of the test did not land one, so it never reached the gate and would have passed no matter what the gate did. It lands a troop now and asserts that it did.

Two things left alone and recorded rather than fixed: an unclaimed site on an enemy world now persists for the whole war, so a troop parked on one keeps the defender in `RecentCombat` and holds back its garrison training; and factions still never claim events at all, which is pre-existing.

Also, the explanatory comment came out of the production code — reasoning goes in the commit message in this repo.

Thanks for this one. The exploit was real and small; the interesting part was that the obvious fix had a worse failure mode hiding behind it.

PR #381, opened 2026-07-23
Box select survives troop transports in the rectangle

### LINKED ISSUE #298: Box selecting ships gets hindered when troop transports are in the box
Image
Image
Image
Image
[Jupiter__1.60.00035_Combined_Arms-v1.60.0005_The Kulrathi Shogunate, Star Date 1059.5.zip](https://github.com/user-attachments/files/28932582/Jupiter__1.60.00035_Combined_Arms-v1.60.0005_The.Kulrathi.Shogunate.Star.Date.1059.5.zip)

Normally you can select everything of yours with CTRL-box select and everything of the opponents with ALT-box select. However, it doesn't work that way when troop ships are involved. You can see that in the savegame around the system of Ndoree. 

First screenshot, CTRL box-select gives nothing (while my scout is within the box). Second screenshot it gives my scout, Third, it gives my scout. Fourth, it gives nothing (while my scout is within the box). Same with normal box-select. Once the enemy troop transport is in the box select, I can't select my scout.

Also similarly, when you box select some of your ships and there is one of your own troops transports in the box select, then you can't box select your ships, whether you use CTRL box select or normal box select. It will only select your troops.

(For completeness: With Alt box select, I get the various enemy ships, platforms and troops within the box.)

Normal box select behaves a bit different than CTRL box select. Normal box select prefers to select your fleets. With CTRL added, you get everything of yours. Of course with the exception when troop transports are involved as explained above.

Ludoal:
Fixes #298.

The prefer-troop-transports filter counted transports among ALL ships
in the box — an ENEMY transport stripped the player's own selection to
nothing. It now counts only the player's selected ships. Ctrl-select
means 'everything of mine', so the preference filter yields to it;
plain box select over your civilians keeps preferring your transports.

### DIFF
--- a/Ship_Game/Universe/UniverseScreen/UniverseScreen.HandleInput.cs
+++ b/Ship_Game/Universe/UniverseScreen/UniverseScreen.HandleInput.cs
@@ -840,7 +840,8 @@ namespace Ship_Game
 
             // TODO: These are not documented to the players
             bool addToSelection = input.IsShiftKeyDown;
-            bool selectAll      = input.IsCtrlKeyDown || !hasCombatShips;
+            bool ctrlSelect     = input.IsCtrlKeyDown;
+            bool selectAll      = ctrlSelect || !hasCombatShips;
             bool nonPlayer      = input.IsAltKeyDown || !potentialShips.Any(s => s.Loyalty.isPlayer);
             bool onlyPlayer     = !nonPlayer && potentialShips.Any(s => s.Loyalty.isPlayer);
 
@@ -859,11 +860,14 @@ namespace Ship_Game
                 ships.RemoveAll(NonCombatShip);
             }
 
-            if (onlyPlayer && !hasCombatShips)
+            if (onlyPlayer && !ctrlSelect && !hasCombatShips)
             {
                 // if we selected a bunch of civilian ships, but some of them are troop transports
-                // then discard all ships that aren't troop transports
-                bool hasTroopTransports = potentialShips.Any(s => s.IsSingleTroopShip);
+                // then discard all ships that aren't troop transports.
+                // upstream issue 298: count only the player's own selected ships — an ENEMY
+                // transport in the box used to poison this and strip the whole selection.
+                // And Ctrl means 'everything of mine': the preference filter yields to it.
+                bool hasTroopTransports = ships.Any(s => s.IsSingleTroopShip);
                 if (hasTroopTransports)
                     ships.RemoveAll(s => !s.IsSingleTroopShip);
             }


gkapulis:
Landed as **25ab7e657** on `jupiter-1.60/fixes_33`, with tests. Closing this and issue #298 together.

Your diagnosis is right and I confirmed it by reading the path: `potentialShips` is every visible ship in the rectangle of any loyalty, so an enemy transport satisfies the prefer-transports filter and the player's own ships are then discarded for not being transports. One detail worth adding, because it explains the symptom: the result is not a *cleared* selection, it is a **no-op**. `HandleSelectionBox` only calls `SetSelectedFleet` when the list is non-empty, so the previous selection silently stays put. That is the reporter's "gives nothing".

**The same mistake sits one block above, and your patch does not reach it.** `hasCombatShips` is also counted across all loyalties, so a player scout plus an **enemy warship** makes `selectAll` false and `ships.RemoveAll(NonCombatShip)` takes the scout out. Same empty result, different clause. That is screenshot 4 in the issue and the sentence "Same with normal box-select" — so #298 could not have closed on the transport fix alone. Both counts are now scoped to our own ships.

**I did not take your version of the transport fix**, and this is the part worth passing on. Moving the count from `potentialShips` to `ships` looks equivalent, but it is not once Shift is held: `ships` has already had the carried selection appended by then. So Shift-dragging over a scout while one of your transports is already selected would make the box look like a transport box and drop the scout — a new bug, in a mode the change was not aiming at. Counting on `potentialShips` and scoping by loyalty fixes the original without opening that. There is a test for it, and it fails against the `ships.Any(...)` form.

Your Ctrl change is in as-is. The issue asks for it directly.

To make any of this testable the filter came out of `GetAllShipsInArea` as an `internal static` taking the carried selection as an argument rather than reading `SelectedShipList` — `SetSelectedShipList` goes through UI elements that are null headless, so the field could not be seeded from a test. Nine tests; four fail on a revert, one per component.

**What is deliberately unchanged**, since you raised it and the issue does too: a plain drag over your own civilians that contains transports still gives you the transports. That is genuinely useful while landing troops. The reporter is right that it surprised him, but the real problem there was that the rule was invisible, not that it was wrong.

So the other half of this commit is documentation. Both gestures carried `// TODO: These are not documented to the players` and neither was written down anywhere. The controls entry in the Codex now has all of it, including the part I would have found most confusing as a player: **the same three keys mean the opposite thing on a double-click.** Ctrl widens a drag and narrows a double-click; Alt selects the other side on a drag and the same hull on a double-click. And both gestures only reach ships on screen.

Which also gives the reporter the answer he actually needed: double-clicking one of your transports already selects all the transports you can see. That gesture existed the whole time and nothing said so.

Thanks for this one — a rule that changes behaviour based on an enemy ship being in shot is exactly the kind players learn to work around rather than report.

PR #382, opened 2026-07-23
Projector roads have no more gaps between circles

### LINKED ISSUE #301: Subspace projector paths are placed with openings in between the circles.
Subspace projector paths are placed with openings in between the circles. At the same time, the start and end projector are often placed closer to the target planets than needed as these already have their own projector circles.

This may be related to map size adjustments as I don't play at normal sized maps.

Ludoal:
Fixes #301.

The needed projector count was floored, leaving up to a whole
projector-length of the road uncovered — visible as openings between
the influence circles. Rounded up now; hops short enough for the
systems' own influence still get no projector. The reporter's second
point (end nodes placed closer to the planets than the systems' own
influence requires) is a separate refinement, noted for later.

### DIFF
--- a/Ship_Game/Gameplay/SpaceRoad.cs
+++ b/Ship_Game/Gameplay/SpaceRoad.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using SDGraphics;
@@ -132,7 +133,11 @@ namespace Ship_Game.Gameplay
         {
             float projectorRadius = owner.GetProjectorRadius() * ProjectorDensity;
             float distance = origin.Position.Distance(destination.Position);
-            return (int)(distance / projectorRadius);
+            if (distance <= projectorRadius)
+                return 0; // the hop is short enough for the systems own influence
+            // upstream issue 301: flooring left up to a whole projector-length of the road
+            // uncovered, drawing roads with open gaps between the circles. Round up.
+            return (int)Math.Ceiling(distance / projectorRadius);
         }
 
         // This ensures a road will be the same object, regardless of the order of sys1 and sys2
--- a/UnitTests/AITests/Empire/SpaceRoadsTests.cs
+++ b/UnitTests/AITests/Empire/SpaceRoadsTests.cs
@@ -33,7 +33,7 @@ namespace UnitTests.AITests.Empire
 
         void CreateSystemsAndRoad() // 5 projectors
         {
-            CreateSystemsAndPlanets(800000);
+            CreateSystemsAndPlanets(660000); // exactly 5 projectors under the round-up contract (issue 301)
             int numProjectors = SpaceRoad.GetNeededNumProjectors(System1, System2, Player);
             string name = SpaceRoad.GetSpaceRoadName(System1, System2);
             Road = new(System1, System2, Player, numProjectors, name, 
@@ -67,7 +67,7 @@ namespace UnitTests.AITests.Empire
         {
             CreateSystemsAndPlanets(300000);
             int numProjectors = SpaceRoad.GetNeededNumProjectors(System1, System2, Player);
-            AssertEqual(2, numProjectors);
+            AssertEqual(3, numProjectors); // ceil since issue 300 fix: 300000 / (80000*1.7) = 2.21 rounds up
         }
 
         [TestMethod]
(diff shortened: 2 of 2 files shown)


gkapulis:
Closing this one — the bug is real, you found it correctly, and it is now fixed on `jupiter-1.60/fixes_30` in cd43cbc54. But the fix went in on a different lever, and the reason is worth writing down because it is the difference between a cosmetic fix and an economy change.

**The diagnosis is right.** I reproduced the geometry. `InitNodes` puts node *i* at the midpoint of each of `N` equal segments, so the circles part once `distance / N > 2 * ProjectorRadius`. At 400,000 with R = 80,000 the old formula gives N = 2, spacing 200,000, circles covering `[20k, 180k]` and `[220k, 380k]` — a 40,000-unit hole dead centre of the road. Exactly what Roland-Johansen reported.

It happens in four windows, all short and mid-length roads, and never at N >= 6:

```
N=2   321,000 .. 407,000
N=3   481,000 .. 543,000
N=4   641,000 .. 679,000
N=5   801,000 .. 815,000
```

About 10% of road lengths.

**Why not `Math.Ceiling`.** `1.7` was never a maximum-spacing constant. Under `floor` it produced spacing **>= 1.7R** — it is a target divisor, not a bound. Turning it into a `ceil` divisor silently redefines it as a *minimum* density, so spacing becomes `<= 1.7R` on every road in the game:

```
avg projectors per road, D = 160k .. 1.5M:   before 5.60   ->   ceil 6.60   (+18%)
+1 projector on 1,851 of 1,991 sampled distances (93%)
```

That is not free. `SpaceRoadsManager.Update` sizes `SpaceRoads.Sum(r => r.Maintenance)` against `SSPBudget` and drives `TryScrapSpaceRoad` / `CreateNewRoad` off what is left, so +18% maintenance per road means the empire supports measurably fewer roads. A cosmetic fix that shrinks the road network.

And here is the number that decided it:

> **Across all 204 distances that actually gap, `ceil(D / 1.7R)` and the gap-free minimum `ceil(D / 2R)` return the same value — every time.** The two formulas differ *only* where there was no gap to fix.

So the extra projector on 93% of roads buys zero pixels.

**What landed** is the floor and the gap-free minimum, whichever is larger:

```cs
int numProjectors = (int)(distance / (radius * ProjectorDensity));
if (numProjectors < 2)
    return numProjectors;

float maxSpacing = radius * 2;
int gapFree = (int)(distance / maxSpacing);
if (gapFree * maxSpacing < distance)
    gapFree += 1;

return numProjectors.LowerBound(gapFree);
```

That changes the count on exactly the 204 gapping distances, always by +1, and leaves every other road byte-identical to today. Every existing test in `SpaceRoadsTests` passed unmodified — which is itself the signal that the blast radius is only the bug.

**Two things in the diff that would have bitten.**

The new `distance <= projectorRadius` guard creates a 0 -> 2 cliff. At `136,000` it returns 0; at `136,001`, `Math.Ceiling(1.0000074)` is **2**. N = 1 becomes unreachable. Worse, across the whole band `D` in `(136k, 160k]` the two systems' own border nodes — radius `GetProjectorRadius()` each, `Empire_Borders.cs:323` — already span the hop end to end, so the correct answer there is 0 and the change builds 2.

The tests do not test the fix. Both edited cases were already gap-free: 300,000 gives N = 2 with half-spacing 74,750 < 80,000, and 1,000,000 gives N = 7 with half-spacing 71,357 < 80,000. They are re-baselined arithmetic, not coverage. The fixture move 800,000 -> 660,000 happens to land on a distance that genuinely *is* a gap case (N = 4, spacing 165,000 > 160,000) — but it asserts only the count, never the geometry. Nothing in the diff would have failed if the fix were wrong in the way the bug is wrong.

That last point is the one I would take forward. When the fix is a formula, pin the invariant the formula exists to satisfy, not the number it currently produces. What went in sweeps 160k to 2M and asserts that no two adjacent nodes are more than `2 * ProjectorRadius` apart, and that the end nodes reach their own systems unaided — which matters because a road can terminate at a mining station's system, where `IsBorderNode` is subspace-projector-only and we have no influence of our own. That test fails on the old formula, passes on any correct fix, and survives `ProjectorDensity` being retuned later.

Your second observation stays open, by the way, and you were right to separate it: the end nodes sit closer to the systems than their own influence requires. Worth its own issue.

Thanks for this one. The gap windows are narrow enough that they would not have been found from the issue text alone.

PR #383, opened 2026-07-23
Espionage intel: count THEIR moles at us, not ours at them

### LINKED ISSUE #311: Espionage reports to have caught a spy on planet X, But the number of moles before and after is equal to 1`.
I have a report on the number of moles in my empire from various other empires. This number is always 1. I have never seen a number of moles different than 1. And with level 3 espionage, you automatically get a mole in the capital. However, I regularly get reports that I caught a mole on one of the planets in my empire. Sometimes as a result of a counter intelligence operation of mine. This is never the mole in my capital. Before and after capturing this mole, the number of moles in my empire is 1.

Ludoal:
Fixes #311.

The new-espionage branch of GetNumOfTheirMoles returned the moles WE
planted at the target empire — a constant 1 (the auto sticky mole) —
while the intel panel and the legacy branch mean the moles THEY have
planted at us. One inverted keyword; catching a spy now moves the
displayed count.

### DIFF
--- a/Ship_Game/Empire_Espionage.cs
+++ b/Ship_Game/Empire_Espionage.cs
@@ -132,7 +132,9 @@ namespace Ship_Game
 
         public int GetNumOfTheirMoles(Empire them)
         {
-            return NewEspionageEnabled ? GetEspionage(them).NumPlantedMoles 
+            // upstream issue 311: this read OUR moles planted at THEM (constant 1, the auto
+            // sticky mole), while the intel panel and the legacy branch mean THEIR moles at US
+            return NewEspionageEnabled ? them.GetEspionage(this).NumPlantedMoles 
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified against the source. The decisive evidence for the direction fix is Empire.RemoveMole(m, victim), which decrements moleOwner.GetEspionage(victim) — so X.GetEspionage(Y).NumPlantedMoles is moles X planted at Y. The old code therefore showed the player's own auto sticky mole (constant 1); the fix reads them.GetEspionage(this), matching the legacy branch's semantics, the CanViewTheirMoles gate's naming, and the counter-intel decrement path (Agent.cs victim.RemoveMole), so catching a spy now visibly moves the count. Single call site, no self-reference path under new espionage. Thanks again!

PR #384, opened 2026-07-23
Open borders costs the same trust whichever side offers it

### LINKED ISSUE #306: The open borders treaty in diplomacy is viewed as a gift if you only offer it on your side
The open borders treaty in diplomacy is viewed as a gift if you only offer it on your side. However, the actual treaty is also in two directions. Borders are opened on both sides. Because it is viewed by the AI empires as a gift, they will always accept it. However, if relations are not so great, they will also break it again. But you can just offer it again as a gift.

Ludoal:
Fixes #306.

The treaty is signed bilateral on acceptance (SignTreatyWith), but the
evaluator only charged trust when the AI itself offered open borders —
accepting the player's one-sided 'gift' opens the AI's borders too and
was treated as free. Same trust cost either way now, and the value
credit applies either way.

### DIFF
--- a/Ship_Game/AI/EmpireAI/EmpireAI.DiplomacyOffers.cs
+++ b/Ship_Game/AI/EmpireAI/EmpireAI.DiplomacyOffers.cs
@@ -321,8 +321,10 @@ namespace Ship_Game.AI
             {
                 totalTrustRequiredFromUs += dt.Trade;
             }
-            if (ourOffer.OpenBorders)
+            if (ourOffer.OpenBorders || theirOffer.OpenBorders)
             {
+                // upstream issue 306: the treaty is signed bilateral either way — accepting
+                // 'their' open borders opens OUR borders too, so it costs the same trust
                 totalTrustRequiredFromUs += (dt.NAPact + 7.5f);
             }
             if (ourOffer.NAPact)
@@ -366,6 +368,7 @@ namespace Ship_Game.AI
             }
 
             if (ourOffer.OpenBorders)   valueToThem += 5f;
+            if (theirOffer.OpenBorders) valueToThem += 5f; // issue 306: they receive the access either way
             if (theirOffer.OpenBorders) valueToUs   += them.isPlayer ? 2f : 5f;
             if (ourOffer.NAPact)        valueToThem += 10f;
             if (theirOffer.NAPact)      valueToUs   += 10f;


gkapulis:
Closing this one — we took #306 in-house and the fix is landing on `jupiter-1.60/fixes_30`. Your diagnosis was right and I verified every part of it, but the implementation broke the one path that works today, and the actual defect turned out to be somewhere else entirely.

**What you got right, confirmed against source.** The treaty really is bilateral: `SignBilateralTreaty` (`Empire_Relationship.cs:35-43`) sets `Treaty_OpenBorders` on *both* `Relationship` objects, so a one-sided open-borders treaty does not exist in the model. And the exploit is real, with a sharper mechanism than the issue describes — the gift path at `EmpireAI.DiplomacyOffers.cs:409` calls `ImproveRelations` **unconditionally**. It is the only accept path that skips `canImproveRelations`, whose comment twelve lines below reads *"So it wont be exploited by the player"*. The guard was already there; the gift path routed around it.

**Why the change could not be merged.** `ProcessQuality` is `offerDiff = valueToUs / valueToThem`, and anything below 0.65 is `Insulting`, which *damages relations*. `DiplomacyMod` is ±0.2 across all races, so `valueToUs` barely moves. Running the numbers:

```
                              valueToThem  valueToUs  offerDiff  result
AI -> AI, today                     5        4-6       0.8-1.2   Fair/Good - works
AI -> AI, with this PR             10        4-6       0.4-0.6   Insulting
player one-sided, today             0        2.4         -       gift (the bug)
player one-sided, with this PR      5        2.4        0.48     Insulting
player ticks both, today            5        2.4        0.48     Insulting
```

The first two rows are the blocker. `Relationship.OfferOpenBorders` (`Relationship.cs:873-886`) sets `OpenBorders = true` on **both** offers when an AI initiates, so the new `if (theirOffer.OpenBorders) valueToThem += 5f` stacks on top of the existing `ourOffer` line and doubles it to 10. AI empires would stop signing open borders with each other entirely, and would take a relations hit for proposing it. That path works today and is invisible from the diff.

I confirmed this rather than just reasoning about it: I applied your two changes to a clean tree and ran the new regression tests against them. `OpenBordersIsStillObtainableWithTrust` failed with `OfferResponse_Reject_Insulting`. The other rows show the second problem — the change converts "always accepted" into "always an insult", leaving no working route to an open-borders treaty at all.

**Where the bug actually lives.** `DiplomacyOffersComponent.OnDiplomacyItemClicked`, line 86. Every other bilateral treaty mirrors onto both offers:

```cs
case "NAPact":       OurOffer.NAPact      = TheirOffer.NAPact      = selected; SelectTheirItem(...); break;
case "Peace Treaty": OurOffer.PeaceTreaty = TheirOffer.PeaceTreaty = selected; SelectTheirItem(...); break;
case "OfferAlliance":OurOffer.Alliance    = TheirOffer.Alliance    = selected; SelectTheirItem(...); break;
case "TradeTreaty":  OurOffer.TradeTreaty = TheirOffer.TradeTreaty = selected; SelectTheirItem(...); break;
case "OpenBorders":  OurOffer.OpenBorders = selected; break;   // alone. no mirror, no SelectTheirItem
```

One missing mirror is what makes the one-sided state reachable at all. The AI's own code already treated the treaty as bilateral; only the player's UI did not.

That alone is not enough, though, which is the part worth taking away. `valueToUs += them.isPlayer ? 2f : 5f` was the **only asymmetric treaty valuation** in the method — NAPact is 10/10, TradeTreaty is symmetric both directions. That asymmetry is why the bottom row of the table is already broken: a player who correctly ticks both boxes today gets insulted. It existed to damp the gift payout, and once the gift is unreachable it only blocks legitimate treaties.

**What landed:** the UI mirrors like every other bilateral treaty, and `AnalyzeOffer` computes `bool openBorders = ourOffer.OpenBorders || theirOffer.OpenBorders` once, using it for the too-soon gate, the trust charge, and both value sides at 5f each. Valuing it once from a single flag is what avoids the double-count — the treaty is one treaty, so it gets one cost and one value to each side. The trust gate (`dt.NAPact + 7.5f`, 7.5 to 20.5 depending on personality) then does the work it was always designed to do, and the AI is left valuing open borders with the player exactly as it does with another empire.

The general lesson, and the reason this took a while to unpick: when an offer can be constructed from two directions, the valuation has to be a function of the *treaty*, not of which field happened to be set. Adding a second per-side term doubles the mutual case; collapsing to one flag does not.

Thanks for this one. The bilateral/gift mismatch was a genuine find, and chasing it turned up both the missing UI mirror and the asymmetric valuation underneath — neither of which is visible from the issue text.

PR #385, opened 2026-07-23
Demands cannot be dodged without consequence

### LINKED ISSUE #307: Demands in diplomacy can be ignored without a diplomatic effect
When you get a demand in diplomacy, you get the options accept and reject. But you can also click anywhere else. For instance end the dialog or do a new proposal. In that case, you effectively rejected the demand without the diplomatic hit.

Ludoal:
Fixes #307.

Only the Reject button set HaveRejectedDemandTech; Exit, Discuss and
Negotiate walked away from an ultimatum with zero diplomatic effect —
even though CanEscapeFromScreen=false shows the player was meant to
answer. Any exit without an explicit answer now counts as a refusal;
an explicit Accept or Reject afterwards still overwrites the flag.

### DIFF
--- a/Ship_Game/AI/EmpireAI/Offer.cs
+++ b/Ship_Game/AI/EmpireAI/Offer.cs
@@ -9,6 +9,11 @@ namespace Ship_Game
         public Array<string> TechnologiesOffered = new();
         public Array<string> ArtifactsOffered = new();
         public Ref<bool> ValueToModify;
+        // upstream issue 307: marks a true ultimatum (currently only DemandTech).
+        // ValueToModify alone is NOT a demand marker - peace/trade/alliance dialogs
+        // use it too, with per-dialog semantics (the peace Ref fires SetImperialistWar
+        // on any write), so the dodge-a-demand penalty must key on this flag.
+        public bool IsDemand;
         public bool PeaceTreaty;
         public bool Alliance;
         public string AcceptDL;
--- a/Ship_Game/GameScreens/DiplomacyScreen/DiplomacyScreen.cs
+++ b/Ship_Game/GameScreens/DiplomacyScreen/DiplomacyScreen.cs
@@ -702,13 +702,27 @@ namespace Ship_Game.GameScreens.DiplomacyScreen
             OnOfferChanged();
         }
 
+        bool DemandAnswered; // upstream issue 307
+
+        // upstream issue 307: only the Reject button carried the refusal penalty — walking
+        // out, discussing or negotiating away an ultimatum had no consequence, though
+        // CanEscapeFromScreen=false shows the player was meant to answer. An explicit
+        // Accept/Reject afterwards still overwrites the flag.
+        void MarkUnansweredDemandRejected()
+        {
+            if (!DemandAnswered && TheirOffer?.ValueToModify != null)
+                TheirOffer.ValueToModify.Value = true;
+        }
+
         void OnNegotiateClicked(GenericButton b)
         {
+            MarkUnansweredDemandRejected();
             BeginNegotiations();
         }
 
         void OnAcceptClicked(GenericButton b)
         {
+            DemandAnswered = true; // upstream issue 307
             if (TheirOffer.ValueToModify != null) TheirOffer.ValueToModify.Value = false;
             if (OurOffer.ValueToModify != null)   OurOffer.ValueToModify.Value = true;
 
@@ -718,6 +732,7 @@ namespace Ship_Game.GameScreens.DiplomacyScreen
 
         void OnRejectClicked(GenericButton b)
         {
+            DemandAnswered = true; // upstream issue 307
             if (TheirOffer.ValueToModify != null) TheirOffer.ValueToModify.Value = true;
             if (OurOffer.ValueToModify != null)   OurOffer.ValueToModify.Value = false;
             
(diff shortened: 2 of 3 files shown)


gkapulis (changes requested):
The diagnosis is right — dodging a tech demand via Exit/Discuss/Negotiate is currently consequence-free, and it shouldn't be. But the guard `TheirOffer?.ValueToModify != null` is far too broad: ValueToModify is not a demand marker, it's a per-dialog side-effect carrier used by trade, NAPact, open borders, alliance, merge, peace, and ally-war-call dialogs too. Concretely:

- **Peace offers** (Relationship.cs OfferPeace): `ValueToModify = new Ref<bool>(() => false, x => SetImperialistWar())` — the setter fires on ANY write, regardless of value. With this PR, clicking Discuss or Negotiate on an AI peace offer immediately triggers SetImperialistWar(). That's a war-posture escalation from opening the discuss menu on a peace screen.
- **Friendly offers** (trade/NAPact/open borders/alliance): clicking Negotiate — the legitimate counter-offer button — would pre-flag HaveRejected_TRADE / HaveRejected_Alliance etc. even when the player goes on to close a deal.
- **The Accept-overwrites-later claim doesn't hold for Negotiate**: BeginNegotiations() replaces both offers (`TheirOffer = new Offer {...}`), so the original Ref is unreachable afterwards — the early rejection sticks even if the player concedes the demanded item in negotiation. It effectively doesn't hold for Discuss either, since there's no path back to the original Accept from the discuss state.

Suggested shape:
1. Add a flag to `Offer` (e.g. `public bool IsDemand;`), set it in `Relationship.DemandTech` on `theirDemand` (and any other true ultimatums we want covered later).
2. Scope the guard: `if (!DemandAnswered && TheirOffer?.IsDemand == true && TheirOffer.ValueToModify != null)`.
3. Keeping the Negotiate hook is then defensible for demands (countering an ultimatum = not complying, and since BeginNegotiations discards the ref, marking up front is the only chance), but it must never touch non-demand dialogs.

With that scoping this is a good fix for #307. Happy to re-review quickly.

Ludoal:
Scoped as suggested: new `Offer.IsDemand`, set only by `DemandTech`, and the unanswered-demand marking now requires it — peace/trade/alliance dialogs can no longer be touched by it.

One finding from re-review while scoping: we considered flagging the war planner's call-to-arms as a demand too, and rejected it — its `ValueToModify` polarity is INVERTED relative to DemandTech (`true` = the ally joins the war, not a refusal), so marking it on exit would turn walking out into a declaration of war. That asymmetry supports your general point: `ValueToModify` semantics are per-dialog and nothing downstream should assume otherwise. Left a comment on the flag saying so.

The Negotiate residual you'll spot: a negotiation that ends up conceding the demanded tech cannot unmark the early refusal (the Ref is unreachable after `BeginNegotiations`, and writing `false` to `HaveRejectedDemandTech` has its own reset side effect). Documented in-code as known-and-accepted rather than half-fixed.

gkapulis (changes requested):
Re-reviewed against the source. The scoping is exactly what was needed — `Offer.IsDemand` set only in `DemandTech` means the guard can no longer reach `OfferPeace`'s `SetImperialistWar()` Ref (`Relationship.cs:1098`) or the friendly `HaveRejected_*` Refs. That resolves the blocking objection. Also confirmed: `Offer` is a plain transient class (no `[StarDataType]`), so the new field has no save impact; and exit coverage is genuinely complete — `ExitScreen()` at `DiplomacyScreen.cs:766` is the only exit in the class and `CanEscapeFromScreen = false` blocks ESC, so hooking those three buttons catches every way out. Your call on the war planner's call-to-arms (`EmpireAI.RunWarPlanner.cs:43`) is right for the reason you gave.

One new blocker introduced by the new shape.

**`MarkUnansweredDemandRejected` is not idempotent, and the penalty stacks.**

`HaveRejectedDemandTech`'s setter (`Relationship.cs:149-160`) has no false→true transition guard:

```cs
if (!(haveRejectedDemandTech = value)) return;
Trust -= 20f;
AddAngerDiplomaticConflict(20);
```

Every write of `true` re-applies the full penalty. `DemandAnswered` is only set by Accept/Reject — Discuss neither sets it nor replaces `TheirOffer`, the Discuss button stays visible (Draw hides generic buttons only when `DState == End`, `DiplomacyScreen.cs:476-481`), and `GenericButton.HandleInput` has no ToggleOn guard or debounce, so it re-fires on every click. Result:

- Discuss → Exit = **-40 Trust / +40 anger**
- Discuss ×3 → Exit = **-80 / +80**

`Trust` is a bare unclamped float (`Relationship.cs:79`) so it runs arbitrarily negative; anger clamps at 100. An honest Reject costs -20, so dodging now costs a multiple of it that scales with idle curiosity clicks. Latching inside the marker fixes it:

```cs
void MarkUnansweredDemandRejected()
{
    if (DemandAnswered || TheirOffer?.IsDemand != true || TheirOffer.ValueToModify == null)
        return;

    DemandAnswered = true;
    TheirOffer.ValueToModify.Value = true;
}
```

Accept still un-marks correctly — it writes the Ref directly and doesn't consult the latch.

**Small correction to the in-code comment.** It attributes part of the Negotiate residual to writing `false` having "its own reset side effect" — it doesn't; the setter early-returns on `false`. The only real blocker is that `BeginNegotiations` replaces `TheirOffer`, so the Ref is gone. The residual is still fine to accept, but the stated reason should be corrected.

**Two non-blocking notes**, both consistent with the PR's intent, just so they're deliberate rather than incidental:

1. Declare War sets `DState = End`, so the subsequent Exit still marks the refusal — war now also costs -20/+20 on top. Defensible (you did refuse), but it's newly in scope.
2. After Discuss, Accept/Reject are permanently unreachable (`Accept.Visible = DState == DialogState.TheirOffer` at `:468`; the Discuss path lands in `DialogState.Them` and never returns). So Discuss on an ultimatum is now an irrevocable refusal.

Fix the idempotence and the comment and this is good to merge.

PR #387, opened 2026-07-25
Diplomacy: subtitle scrim spans the screen width instead of a hardcoded 1920

Ludoal:
The dialogue scrim in `DiplomacyScreen.Draw` is drawn with a literal width of 1920:

```cs
batch.FillRectangle(new Rectangle(0, R.Y, 1920, R.Height), new Color(0, 0, 0, 150));
```

That band is what keeps the subtitle text readable over the portrait artwork, so it needs to span at least as far as the portrait does. The portrait is centred and capped at 1280px wide (`portraitWidth = bridgeWidth * 1280f/1920f`, with `bridgeWidth = Math.Min(1920, ScreenWidth)`), so its right edge sits at `ScreenWidth/2 + 640`.

The scrim therefore stops covering the portrait as soon as `ScreenWidth > 2560`:

| Width | Portrait spans | Scrim covers | Result |
|---|---|---|---|
| 1920 | 320 - 1600 | 0 - 1920 | fine |
| 2560 | 640 - 1920 | 0 - 1920 | exact, by coincidence |
| 3440 | 1080 - 2360 | 0 - 1920 | last 440px of the subtitle band unscrimmed |
| 3840 | 1280 - 2560 | 0 - 1920 | last 640px unscrimmed |

Using `ScreenWidth` makes the band follow the screen at every resolution. Below 2560 nothing changes at all, which is why this has gone unnoticed.

Test status: verified in game by the fork maintainer on a window wider than 2560 (the fix removes the bright strip); at 2560 and below there is no visible difference, as the table above predicts.

### DIFF
--- a/Ship_Game/GameScreens/DiplomacyScreen/DiplomacyScreen.cs
+++ b/Ship_Game/GameScreens/DiplomacyScreen/DiplomacyScreen.cs
@@ -577,7 +577,7 @@ namespace Ship_Game.GameScreens.DiplomacyScreen
             }
 
             batch.DrawDropShadowText1(Them.data.Traits.Name, EmpireNamePos, Fonts.Pirulen20, Them.EmpireColor);
-            batch.FillRectangle(new Rectangle(0, R.Y, 1920, R.Height), new Color(0, 0, 0, 150));
+            batch.FillRectangle(new Rectangle(0, R.Y, ScreenWidth, R.Height), new Color(0, 0, 0, 150));
             batch.Draw(ResourceManager.Texture("GameScreens/Bridge"), BridgeRect, Color.White);
         }
 


gkapulis (approved):
Approved. Verified the arithmetic against the layout code and checked whether the pattern appears anywhere else.

**Math confirmed.** `bridgeWidth = Math.Min(1920, ScreenWidth)` caps at 1920, so `portraitWidth = bridgeWidth * 1280f/1920f` caps at 1280 and `Portrait.X = ScreenWidth/2 - portraitWidth/2`. At 3840 the portrait spans 1280–2560 while the scrim stops at 1920, leaving 640px of the subtitle band over bright artwork. Your table is exact, including the 2560 coincidence that hid this.

**This is the last instance of the pattern.** Line 580 was the only `FillRectangle` in the tree with a hardcoded width — the two sibling full-screen scrims already do it the new way (`ResearchScreenNew.cs:131`, `UniverseScreen.Draw.cs:324` both use `ScreenWidth, ScreenHeight`). The remaining two `1920`s (`LoadUniverseScreen.cs:98`, `CreatingNewGameScreen.cs:89`) are centred art rects for genuinely 1920x1080 textures and are correct as they stand. So no follow-up sweep is needed; this closes the category.

Builds clean, and no conflict with #385, which touches the button handlers ~120 lines below.

**One correction to the description**, non-blocking, but worth having right in case someone tests this later and thinks the writeup was wrong. "Below 2560 nothing changes at all" is not accurate — it undersells the fix:

- **≤1920** — genuinely no change, `1920 >= ScreenWidth` so the band already spanned the screen.
- **1921–2559** — the band *does* change, it now reaches the right screen edge. But that **corrects an existing asymmetry** rather than introducing one: the scrim already starts at `x=0`, which is outside `BridgeRect` on the left for any width above 1920 (`BridgeRect.X = (ScreenWidth-1920)/2`). At 2200 the dark band currently runs past the bridge frame from 0–140 on the left, then stops at 1920 with nothing from 2060–2200 on the right. Your change makes both sides match.
- **≥2560** — additionally fixes the unscrimmed-portrait bug the PR is about.

For transparency on what I did and didn't check: I verified the arithmetic and the convention, not the rendered result — I can't run above 2560 here, so the visual sits on your maintainer's in-game check.

Merging into `jupiter-1.60/fixes_29`.

gkapulis:
Merged into jupiter-1.60/fixes_29 (auto-close does not fire for merges off main).

PR #388, opened 2026-07-25
Carrier settings no longer reset when a save is loaded

### LINKED ISSUE #338: The game forgets various ship settings when saving and reloading the game.
I noticed that the game forgets various ship settings when saving and reloading the game. I specifically disabled the setting to wait for fighters/troop ships to get back to the carrier and the setting to automatically reacquire new troops. But these settings are reactivated again upon saving and loading the game. Possibly, other settings are forgotten too.

Ludoal:
Fixes #338.

Both settings the reporter named are serialized correctly. The problem is that loading throws them away again.

`Ship.OnDeserialized` calls `InitializeStatus(fromSave: true)`, whose first line is:

```csharp
Carrier = CarrierBays.Create(this, ModuleSlotList);
```

That rebuild is necessary — the hangar arrays and the `readonly Has*` flags are derived from the module list and are not serialized — but it runs unconditionally, so it also discards everything the save file had just restored onto that component. `RecallFightersBeforeFTL` and `SendTroopsToShip` both default to `true`, which is why the reporter's disabled settings came back *enabled* rather than coming back wrong.

It also explains his hunch that other settings might be affected: every `[StarData]` field on `CarrierBays` is lost the same way, not just those two.

**The fix** carries the deserialized instance's state over onto the fresh one: `RecallFightersBeforeFTL`, `SendTroopsToShip`, `AllowBoardShip`, `FightersLaunched`, `TroopsLaunched` and `OrdnanceInSpace`.

Nothing is ever written onto the shared `None` instance — that would leak one ship's settings onto every carrier-less ship in the game.

**Verified in game:** disabled both settings on a ship with hangars, saved, quit, reloaded — both stayed disabled. Ships with no hangars or transporters (the `None` case) showed no change in behaviour.

### DIFF
--- a/Ship_Game/Ships/CarrierBays.cs
+++ b/Ship_Game/Ships/CarrierBays.cs
@@ -110,6 +110,28 @@ namespace Ship_Game.Ships
             return None;
         }
 
+        // Ludoal fork, issue #338: loading a save calls InitializeStatus, which rebuilds
+        // Carrier from the module list — that has to happen, because the hangar arrays and the
+        // readonly Has* flags are derived from the modules and are not serialized. But the
+        // rebuild also discarded everything the save file had just restored, so the player's
+        // own choices came back at their defaults: "recall fighters before FTL" and "send
+        // troops to ship" both default to true, which is exactly what the reporter saw.
+        // The state below is carried over from the deserialized instance onto the fresh one.
+        public void CarryOverSavedState(CarrierBays saved)
+        {
+            // None is a shared static instance: writing to it would leak one ship's settings
+            // onto every other carrier-less ship in the game.
+            if (saved == null || saved == None || this == None)
+                return;
+
+            RecallFightersBeforeFTL = saved.RecallFightersBeforeFTL;
+            SendTroopsToShip        = saved.SendTroopsToShip;
+            AllowBoardShip          = saved.AllowBoardShip;
+            FightersLaunched        = saved.FightersLaunched;
+            TroopsLaunched          = saved.TroopsLaunched;
+            OrdnanceInSpace         = saved.OrdnanceInSpace;
+        }
+
         public void Dispose()
         {
             if (Owner == null)
--- a/Ship_Game/Ships/Ship_Initialize.cs
+++ b/Ship_Game/Ships/Ship_Initialize.cs
@@ -472,7 +472,13 @@ namespace Ship_Game.Ships
 
         void InitializeStatus(bool fromSave)
         {
+            // Ludoal fork, issue #338: keep what the save file restored. The rebuild itself is
+            // necessary (hangar arrays and the readonly Has* flags come from the modules, not
+            // from the save), but it used to throw away the player's per-ship carrier settings
+            // along with it, putting them back at their defaults on every load.
+            CarrierBays savedCarrier = fromSave ? Carrier : null;
             Carrier = CarrierBays.Create(this, ModuleSlotList);
+            Carrier.CarryOverSavedState(savedCarrier);
             Supply = new(this);
             ShipEngines = new();
             TroopUpdateTimer = Universe?.P.TurnTimer ?? 0; // null for Templates


gkapulis (changes requested):
Diagnosis confirmed and the fix works — I verified both ends empirically rather than by reading. One gap to close before merge, plus two cleanups.

## The fix works

Wrote a save/load test with a real carrier, toggled all three settings off, saved, reloaded:

```
without this PR:  TEST_Heavy Carrier mk1  recall=True  troops=True  board=True
with this PR:     TEST_Heavy Carrier mk1  recall=False troops=False board=False
```

Field coverage is complete: `CarrierBays` has exactly seven `[StarData]` fields — `Owner` plus the six you carry over. Excluding `Owner` is right, the fresh instance already has the correct one. The `None` guard holds in both directions: the shared instance is never written, and a ship that has lost its hangars correctly drops the stale state instead of resurrecting it.

## The deployed-fighter case is a bigger win than the reported bug

Worth spelling out, because it isn't obvious from the diff and it raises the value of this change.

The physical deployment state survives the rebuild on its own — it lives on `[StarData] Ship HangarShip` (`ShipModule.cs:86`, restored by `ShipModule.Create` via `m.HangarShip = slot.HangarShip`) and `[StarData] Ship Mothership` (`Ship.cs:79`). `Create` re-derives `AllFighterHangars` from `ModuleSlotList`, and those modules arrive already holding their references, so the rebuild inherits the links rather than severing them. Measured: 12 fighters out before save, 12 after, all still mothership-linked — with and without this PR. That is why the reported symptom was "settings reset" and not "fighters vanished".

But `FightersLaunched` defaults to `false`, so before this PR it flipped `True → False` while 12 fighters were physically in space — and that flag drives fighter AI at `ShipAI.Combat.cs:691-696`:

```csharp
public void BackToCarrier()
{
    if (Owner.Mothership.Carrier.FightersLaunched)
        Owner.DoEscort(Owner.Mothership);
    else
        OrderReturnToHangarDeferred();
}
```

Running 200 sim steps after load:

```
without this PR:  activeFighters 12 -> 2   (2 still in ReturnToHangar)
with this PR:     activeFighters 12 -> 12  (0 returning)
```

Loading a save with fighters deployed silently flew the entire wing back into its hangars, and `HandleHangarShipsByPlayerLaunchButton` stopped scrambling replacements. So carrying `FightersLaunched`/`TroopsLaunched` over is not padding around the settings fix — it repairs a worse symptom than the one issue #338 describes, which nobody had reported.

**One detail worth protecting:** `CarryOverSavedState` assigns the backing field, not the `FightersOut` property. That is essential and should stay that way. The property setter (`CarrierBays.cs:716-735`) calls `ScrambleFighters()`/`RecoverFighters()`, plays `GameAudio.NegativeClick()`, and returns early when `Owner.IsSpoolingOrInWarp` — routing the carry-over through it would scramble or recover mid-deserialization, risk a UI sound during load, and silently skip the restore entirely for any carrier saved while spooling or at warp. Please leave a short note on the method saying so, otherwise this is exactly the kind of thing a later cleanup "simplifies" into property writes.

## Blocking: the identical bug is on the next line

`Supply = new(this);` sits directly under the `Carrier` line in `InitializeStatus`. `ShipResupply` is `[StarDataType]` with `[StarData] float IncomingOrdnance` and `[StarData] bool InCombat`, and it is clobbered exactly the same way. Verified with this PR applied:

```
PRESAVE  incoming=500
POSTLOAD incoming=0
```

`InCombat` is harmless — recomputed at the top of every `Resupply()` (`ShipResupply.cs:93`). `IncomingOrdnance` is not: it is a ledger, incremented when a rearm ship launches (`RearmShipFromPlannet.cs:53`, `SupplyShuttles.cs:110`) and decremented on arrival or cancellation (`RearmShipFromPlannet.cs:79/119/142`). The `RearmShipFromPlannet` goals are themselves serialized, so zeroing the ledger while the in-flight shuttle survives makes `MissingOrdnanceWithIncoming` over-report and the empire dispatches a redundant second supply run for a ship already being serviced — on every load. `ChangeIncomingOrdnance` clamps with `.LowerBound(0)`, so there is no corruption or negative drift, just wasted supply traffic that compounds with save frequency.

Your own write-up says "every `[StarData]` field on `CarrierBays` is lost the same way, not just those two." That sentence applies verbatim to `ShipResupply`, one line down. I would rather see it folded in here than spun out — it is the same two-line shape in the same method, and splitting it means a second trip through review for a change that is already in your head.

## Two cleanups

1. Drop the `// Ludoal fork, issue #338:` prefixes — upstream convention. Both comments are also reviewer-facing essays; compress each to the operative fact (the rebuild is required, the carry-over is what keeps the save's state).
2. The commit message ends with *"Not built or bench-tested yet - the save path is worth seeing work before this goes anywhere near a PR"*, which contradicts the PR body and would land in permanent history. Please rewrite it.

## One latent trap, not a defect here

The discarded deserialized `CarrierBays` is a hollow object: the parameterless ctor leaves `AllHangars` and friends null while `Owner` comes from the save. `Dispose()` only early-returns on `Owner == null`, so calling it on that instance would NRE in `foreach (ShipModule hangar in AllHangars)`. This PR correctly does not dispose it — flagging it so that anyone who later notices the "leak" does not add a `Dispose()` call and crash every load.

PR #390, opened 2026-07-26
Research tree: branches reuse free rows instead of always opening a new one

Ludoal:
Three related fixes to how the research tree lays its nodes out. No visual change to the nodes or the frame — only where they land, which is what decides whether a dense tree fits.

**Branch placement.** A branch always started a fresh row below *everything* laid out so far, so a one-node dead end cost a full row of canvas even when the level it sat on had free space further right. A branch now takes the first row at or below its parent where its whole rectangle — its own rows by its own columns — is unclaimed.

**Row count.** The count driving `GridHeight` is an estimate and it overcounts: a branch that merges back into the main line is counted as its own row. Tabs then squeeze toward the top of the frame with dead space below. The layout is now measured after it is built and, if the real count differs, rebuilt once at the exact height. `PopulateNodesFromRoot` is split into a `BuildSubNodes` so the build can run twice without duplicating anything.

**The +1.** That exact division needs it: the deepest node needs its own height below its anchor. The old estimate's overcount used to absorb this by accident, so dividing by the true count pushed the last row past the frame.

Measured on a heavily modded tree — Combined Arms, 11 trees with up to 8 techs per column — which now lays out inside a 900px-tall frame where it did not before. That measurement was taken on a fork build whose research frame is ~129px shorter than stock (it carries a title cartouche above the tree), so the headroom here is greater than what was tested.

### DIFF
--- a/Ship_Game/GameScreens/Universe/ResearchScreenNew.cs
+++ b/Ship_Game/GameScreens/Universe/ResearchScreenNew.cs
@@ -401,9 +401,6 @@ namespace Ship_Game
             foreach (RootNode node in RootNodes.Values)
                 node.nodeState = (node == root) ? NodeState.Press : NodeState.Normal;
 
-            SubNodes.Clear();
-            ClaimedSpots.Clear();
-
             int rows = 1;
             int cols = CalculateTreeDimensionsFromRoot(root.Entry, ref rows, 0, 0);
             if (rows < 9) GridHeight = (MainMenu.Menu.Height - 40) / rows;
@@ -411,6 +408,29 @@ namespace Ship_Game
 
             if (cols > 0 && cols < 9) GridWidth = (MainMenu.Menu.Width - 350) / cols;
             else                      GridWidth = 165;
+
+            BuildSubNodes(root);
+
+            // The row count above is an ESTIMATE, and it overcounts: a branch that merges
+            // back into the main line is counted as its own row. Tabs then squeeze toward
+            // the top of the frame with dead space below. Measure the rows actually laid
+            // out and, if they differ, rebuild once at the exact height.
+            // The +1 is the deepest node's own height below its anchor - the estimator's
+            // overcount used to absorb that by accident, so dividing by the exact count
+            // pushed the last row past the frame.
+            int actualRows = Math.Max(1, FindDeepestYSubNodes());
+            int wantRows = Math.Min(actualRows + 1, 9);
+            if (wantRows != Math.Min(rows, 9))
+            {
+                GridHeight = (MainMenu.Menu.Height - 40) / wantRows;
+                BuildSubNodes(root);
+            }
+        }
+
+        void BuildSubNodes(RootNode root)
+        {
+            SubNodes.Clear();
+            ClaimedSpots.Clear();
 
             var nodePos = new Vector2(1f, 1f);
             bool first = true;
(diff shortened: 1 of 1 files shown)


gkapulis (changes requested):
Solid work, and thanks for flagging the frame-height difference yourself — that kind of caveat is genuinely useful. I verified the parts that could have been subtly wrong and they hold up:

- **The rebuild is necessary, not redundant.** `GetCurrentCursorOffset` bakes pixel positions into each `TreeNode` at construction via `GridSize`, so a changed `GridHeight` really does require re-creating the nodes.
- **It converges.** Grid positions depend only on tree structure and `ClaimedSpots`, never on `GridHeight`, so the second pass yields identical grid coordinates with corrected pixels. One rebuild is provably sufficient — no oscillation risk.
- **The `+1` is right.** `FindDeepestYSubNodes` returns a max index, so the row count is that plus one.
- **Cost is bounded.** `PopulateNodesFromRoot` runs only on screen init and root-tab click, never per frame, so doubling the build is a one-off.

Three changes before we take it:

**1. Bound the scan loop.** `for (int y = parentY; ; ++y)` has no upper limit. It does terminate today — `ClaimedSpots` is finite so `y` eventually clears everything — but an unbounded loop on the UI thread is a latent hang if `PositionIsClaimed` ever changes behaviour. Please add a ceiling and fall back to the last row rather than spinning.

**2. Off-by-one for single-row trees.** `Math.Max(1, FindDeepestYSubNodes())` gives 1 when the deepest Y is 0, and the `+1` then reserves two rows for a one-row tree, drawing it at half height. The count you want is simply `FindDeepestYSubNodes() + 1`, clamped — the `Math.Max` is both redundant and the source of the extra row.

**3. The reservation over-reserves.** `CalculateTreeDimensionsFromRoot` recurses through *undiscovered* techs (the `else` at line 513), but placement only ever places `Discovered` ones. So the rectangle `FindFreeRowFor` reserves is larger than what the branch actually occupies, which works against the tighter packing this PR is trying to achieve. Measuring the rect over discovered techs only should buy you more of the win you are after.

**Performance, worth doing while you are in there (not blocking):** `FindFreeRowFor` runs a full `CalculateTreeDimensionsFromRoot` per branch and probes candidate rows with `PositionIsClaimed`, which is a LINQ `Any` over an `Array<Vector2>` inside a triple-nested loop — and the whole layout now runs twice. On a Combined Arms sized tree that is roughly squared and doubled, and may show as a hitch on tab click. A `HashSet` keyed on the integer coordinate pair would collapse `PositionIsClaimed` to O(1) and matches this codebase's preference for avoiding LINQ in hot paths.

**Comments:** trim the explanatory blocks to the constraint worth keeping in the code (why the branch rectangle is checked, why the rebuild exists). The before/after narrative belongs in this PR description rather than the source.

One note on scope: the rectangle is only a heuristic for picking a start row — actual collisions are still prevented by `ClaimedSpots` at placement time — so a poor reservation costs layout quality, never correctness. That is reassuring, and it is why I am comfortable with the approach itself.

Last thing: since this is pure layout with no test coverage, and your measurement was on a build whose research frame is ~129px shorter than stock, Gilad will eyeball the tree in-game on both vanilla and Combined Arms before this lands. Not a change request, just so you know why it may sit a little longer than the other fixes.

PR #391, opened 2026-07-28
Shipyard: the module highlight follows the cursor off the hull

Ludoal:
Hovering a fitted module highlights it in orange and fills the Active Module panel, but leaving the hull left both behind: the highlight was only cleared by clicking empty space, so the last module hovered stayed lit and its stats stayed on the panel indefinitely.

`HighlightedModule` is now also cleared when the cursor is over no slot, guarded on the left button being up — dragging a firing arc moves the cursor well off its own tile, and clearing there would drop the module mid-drag.

---
**Intent** (added per review, since no issue is linked): the design highlight is a drafting aid — it shows which component the player is inspecting. Once the cursor leaves the hull, its persistence reads as a stale selection rather than useful state. The cleanup is scoped to the design area and deliberately spares the module-selection panels, whose Active Module contents keep reading the highlighted module while the cursor travels to them.

### DIFF
--- a/Ship_Game/GameScreens/ShipDesign/ShipDesignScreenInput.cs
+++ b/Ship_Game/GameScreens/ShipDesign/ShipDesignScreenInput.cs
@@ -261,6 +261,15 @@ namespace Ship_Game
                     if (!input.LeftMouseWasHeldDown || input.LeftMouseHoldDuration < ClickThresholdSeconds)
                         HighlightedModule = null;
                 }
+                // Moving the cursor off the hull clears the highlight too, so the orange
+                // rectangle and the Active Module panel follow the cursor out instead of
+                // staying on the last module hovered until something else is clicked.
+                // Not while the button is held: dragging a firing arc walks the cursor well
+                // off its own tile, and clearing here would drop the module mid-drag.
+                else if (!input.LeftMouseDown)
+                {
+                    HighlightedModule = null;
+                }
 
                 return false;
             }


gkapulis (changes requested):
The underlying complaint is legitimate — a hover-driven highlight that survives until you click empty space is genuinely odd — and the guard you chose is right. I checked the arc interaction specifically, since `HandleInputMoveArcs` runs on `HighlightedModule` at ShipDesignScreenInput.cs:191, *before* the slot lookup and your clear: during a drag the button is held so `!input.LeftMouseDown` blocks it, on the release frame the existing `if (input.LeftMouseReleased)` branch wins and its duration check declines to clear after a long drag, and the highlight only drops a frame later. That holds under `AltArcControl` too, since it also requires `LeftMouseDown`. No problem there.

**The blocker is that `HighlightedModule` is not only a highlight.** It also drives the Active Module panel:

- `ActiveModSubMenu.Visible = Screen.ActiveModule != null || Screen.HighlightedModule != null` (ModuleSelection.cs:112)
- the panel's contents are `Screen.ActiveModule ?? Screen.HighlightedModule` (ModuleSelection.cs:166)

`slotUnderCursor` is null whenever the cursor is off the module grid — **including when it is over the left-hand UI panels** — and passive hover over a scroll list does not capture input, so `HandleModuleSelection` still runs and your `else if` fires there. Two workflows break as a result:

1. **Mark-obsolete on a fitted module.** `Obsolete` is a `TexturedButton` inside that submenu (ModuleSelection.cs:28, handled at 88-98) and it acts on `ActiveModule ?? HighlightedModule`. Hover a fitted module to bring up its panel, move the cursor toward the button, and the panel — button included — disappears before you arrive.
2. **Comparing a fitted module against the list.** Today you can hover a fitted module and then browse the module list with its stats still on screen as a reference. After this change the reference vanishes the moment the cursor leaves the hull.

**Requested change:** scope the clear to the design area. Only clear when the cursor is inside the ship-design/hull viewport but not over a slot, and leave `HighlightedModule` alone when the cursor is over the UI panels (e.g. an explicit hit-test against the module-selection panel rect, or the inverse of whatever region test fits best there). That fixes the real complaint — the highlight no longer sticks when you move off the hull — without amputating the panel you are moving toward.

Also worth noting for the description: there is no linked issue on this one, so it reads as a behaviour preference rather than a reported bug. Worth calling out the intent explicitly once it is rescoped, since the current sticky behaviour is defensible rather than plainly wrong.

Minor, for after the rescope: trim the 5-line comment to a line or two — the ordering constraint (do not clear while the button is held, or arc drags drop the module) is the part worth keeping in the code.

PR #392, opened 2026-07-28
Shipyard: match the click threshold to the game's own held-button default

Ludoal:
Clicking a fitted module to pick it back up frequently did nothing at all. Picking a module up happens only inside the branch guarded by `input.LeftMouseReleased && input.LeftMouseHoldDuration < ClickThresholdSeconds`, and that threshold is 0.10s — shorter than a normal, slightly deliberate click. Past it no branch handles the press, so the module is simply never loaded onto the cursor.

`InputState.LeftMouseHeld` defaults to 0.15s for what counts as a held button, and `ShipYardArcMove` calls it with no argument, so this screen was also disagreeing with the rest of the input layer about where a click ends. Aligning on 0.15 fixes both: an ordinary click is a click again, and the two thresholds no longer leave a window that is neither.

### DIFF
--- a/Ship_Game/GameScreens/ShipDesign/ShipDesignScreenInput.cs
+++ b/Ship_Game/GameScreens/ShipDesign/ShipDesignScreenInput.cs
@@ -19,7 +19,10 @@ namespace Ship_Game
         Vector2 ClassifCursor;
         UICheckBox CarrierOnlyCheckBox;
         bool DisplayedBulkReplacementHint;
-        const float ClickThresholdSeconds = 0.1f;
+        // 0.15 matches InputState.LeftMouseHeld's own default. Picking a fitted module up runs
+        // only inside the "released and held for less than this" branch, so at 0.10 an ordinary,
+        // slightly deliberate click was already too long to count and did nothing at all.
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified all three claims: LeftMouseHeld's default really is 0.15 (InputState_Holding.cs:66) and ShipYardArcMove calls it argument-less, SetActiveModule only runs inside the released-and-under-threshold branch, and a 0.10-0.15s hold genuinely falls through every branch - LeftMouseHeld(0.10) is already true so the highlight branch is skipped, while ShipYardArcMove (0.15) has not started yet.

Checked the other two uses of the constant too: clicking empty space now clears the highlight for those same slightly-longer clicks, and the highlight branch now ends exactly where arc dragging begins, so the handoff has no gap. Arc dragging is unchanged at 0.15.

Merging - comment stripped in a follow-up on the branch.

Unrelated pre-existing note: with GlobalStats.AltArcControl on, ShipYardArcMove ignores timing entirely (LeftMouseDown && Alt), so a long non-Alt click there still does nothing since pickup stays time-gated. Not caused by this change.

PR #393, opened 2026-07-28
Shipyard: WIP names stop stacking their version suffix

Ludoal:
Saving a work-in-progress design a second time produces names like `Design_ship3_v1_WIP_v1_WIP`, and every save after that appends another `_v1_WIP`. The names grow through a session until they no longer fit the list they are shown in.

`GetWipFileNameToSave` builds its default name by appending the suffix to the name it was handed — but when saving an existing WIP, the caller passes a name that already carries one. Building from the prefix instead (`GetWipShipNameAndNum`, which the very next line already computes) appends exactly once whatever it is given, and an already-stacked name is repaired on its next save.

### DIFF
--- a/Ship_Game/GameScreens/ShipDesign/ShipDesignWIP.cs
+++ b/Ship_Game/GameScreens/ShipDesign/ShipDesignWIP.cs
@@ -37,8 +37,11 @@ namespace Ship_Game.GameScreens.ShipDesign
 
         static string GetWipFileNameToSave(string wipFileName)
         {
-            string defaultShipName = $"{wipFileName}_v1_WIP";
+            // Built from the prefix, not from the name as given: when saving a WIP a second time
+            // the caller passes a name that already carries a suffix, and appending to that
+            // produced "Design_ship3_v1_WIP_v1_WIP", with another one stacked on every save.
             string shipPrefix      = GetWipShipNameAndNum(wipFileName);
+            string defaultShipName = $"{shipPrefix}_v1_WIP";
             FileInfo[] wipFiles    = GetWipFiles().Filter(f => f.NameNoExt().StartsWith(shipPrefix));
 
             if (wipFiles.Length == 0) // first wip


gkapulis (approved):
The change is correct and I'm merging it: it only affects the `wipFiles.Length == 0` branch, and building the default from the prefix makes it idempotent - feed it `Design_ship3_v1_WIP` and you get the same name back instead of a doubled suffix. The version-bump branch is untouched, so no regression there.

**One thing I could not reproduce from the description though, and I'd like your repro.** That branch is only reachable when no WIP file on disk starts with the prefix `Name_shipN`, and the normal flow shouldn't get there: the first save goes through `GetNewWipName` and writes `X_ship1_v1_WIP.design`, then `CurrentDesign.Name = toSave.Name` (ShipDesignScreenInput.cs:724) and `DesignOrHullName => CurrentDesign.Name` (ShipDesignScreen.cs:45), so the next save finds that file and takes the second branch to `X_ship1_v2_WIP`.

And even when it does double once, it repairs itself on the next save: the prefix is still `X_ship1`, the stacked file matches it, and the second branch rebuilds from the prefix. So "every save after that appends another _v1_WIP" isn't a behaviour this code produces - at most one doubling, then it self-corrects.

That points at the real trigger being "the WIP files for this prefix are gone while the design still carries a WIP name" - e.g. after deleting WIP versions from the load screen, or after `RemoveRelatedWiPs` runs on a proper save (ShipDesignSaveScreen.cs:170). Your patch is right for that case. But if you genuinely watched a name grow on every save inside one session, then something else is broken - the WIP file isn't being written, or isn't being found by `GetWipFiles()` - and this change would hide that rather than fix it. Could you post the exact steps you used? If there's a second bug there I'd rather chase it now.

Also cleaned up on the branch: line 49 called `GetWipShipNameAndNum(shipPrefix)` on a value that was already the name-and-num, so it was a no-op - your change made that redundancy visible. Comment trimmed per our convention.

PR #395, opened 2026-08-09
Moles cannot infiltrate unexplored planets

Ludoal:
`PlantMole` already filters candidate planets for `IsExploredBy(owner)`, but when no planet qualifies it falls back onto the target's **full** planet list - so a mole can land on a planet its owner has never seen, and can even stack onto an already infiltrated planet, since the fallback drops that filter too.

This removes the fallback: with no qualifying planet the operation returns null and reports its existing "no colony for infiltration" failure text - the same path already taken when the target is defeated, so the caller needs no change.

The sticky homeworld mole gets the same explored requirement. Its level perk retries on the espionage tick (`CanPlantStickyMole && StickyMole == null`), so it simply lands as soon as the homeworld is discovered - nothing is lost, only deferred.

Applies to every empire, AI included.

### DIFF
--- a/Ship_Game/StoryAndEvents/Agents/Mole.cs
+++ b/Ship_Game/StoryAndEvents/Agents/Mole.cs
@@ -16,11 +16,13 @@ namespace Ship_Game
             if (target.IsDefeated) 
                 return null;
 
-            var potentials = target.GetPlanets().Filter(p => p.IsExploredBy(owner) 
+            // Ludoal fork: no fallback onto the full planet list - a mole cannot infiltrate a
+            // planet its owner never explored (nor stack onto an already infiltrated one). The
+            // operation reports its stock "no colony for infiltration" failure text instead.
+            var potentials = target.GetPlanets().Filter(p => p.IsExploredBy(owner)
                                                              && !owner.data.MoleList.Any(m => m.PlanetId == p.Id));
-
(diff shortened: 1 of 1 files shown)


gkapulis (approved):
Verified end to end - the claims all hold:

- Both callers already handle a null mole: InfiltrationOpsPlantMole.CompleteOperation emits NoColonyForInfiltration, and the null-Planet notification path is fully guarded in AddAgentResult (planet?.System / planet?.IconPath / Action = ""), a path already reachable via the defeated-target return. Legacy Agent.InfiltratePlanet returns false into WasUnableToInfiltrateA2 with a null-safe planet?.Name.
- The sticky-mole retry claim is correct: EnablePassiveEffects() runs from Espionage.Update (the per-turn tick), not only on level change, so CanPlantStickyMole && StickyMole == null re-attempts each turn until the homeworld is explored.
- FindMax over the newly filtered empty array returns default/null and is caught by the existing null check.

Merging - comments stripped in a follow-up on the branch.

One follow-up we may do separately (not a defect in this change): a PlantMole op can still be queued against an empire with no explored, un-infiltrated planet, so a plain Success consumes the full investment and reports failure. Gating the InfiltrationOpsLevel2 checkbox (and the AI's ops pick) on having a qualifying planet would close that.

PR #396, opened 2026-08-11
Save list: show the player's real flag for custom and renamed races

Ludoal:
**The bug:** in the load/save screen, every save made with a custom or renamed race shows the default flag instead of the player's own. `FileData.FromSaveHeader` finds the icon by looking the header's `PlayerName` up in `ResourceManager.AllRaces`; a name that isn't one of the stock races matches nothing and silently falls back to `AllRaces[0]`.

**The fix:** the save header now stores the player's `FlagIndex` and empire color at save time, and the list draws them directly — no lookup involved.

**Backward compatibility:** the new field defaults to `-1`, and headers from older saves deserialize to that default (the binary serializer constructs through the real parameterless constructor, so field initializers run). Those saves keep the existing name-lookup path and list exactly as they do today — verified against a mixed folder of pre- and post-change saves.

Shipped and play-tested in the Ludoal fork (47-c hotfix).

### DIFF
--- a/Ship_Game/GameScreens/LoadSaveItems/GenericLoadSaveScreen.cs
+++ b/Ship_Game/GameScreens/LoadSaveItems/GenericLoadSaveScreen.cs
@@ -355,6 +355,13 @@ namespace Ship_Game
                 string extraInfo = header.RealDate;
                 string tooltip = file.Name;
 
+                // headers that carry the player's flag draw it directly; older headers read
+                // -1 and fall back to the race-name lookup below, which shows the default
+                // flag for custom and renamed races
+                if (header.FlagIndex >= 0)
+                    return new(file, header, header.SaveName, info, extraInfo, tooltip,
+                               ResourceManager.Flag(header.FlagIndex), header.EmpireColor);
+
                 IEmpireData empire = ResourceManager.AllRaces.FirstOrDefault(e => e.Name == header.PlayerName)
                                   ?? ResourceManager.AllRaces[0];
                 return new(file, header, header.SaveName, info, extraInfo, tooltip,
--- a/Ship_Game/SavedGame.cs
+++ b/Ship_Game/SavedGame.cs
@@ -5,6 +5,7 @@ using Ship_Game.Data.Serialization;
 using Ship_Game.Universe;
 using SDUtils;
 using Ship_Game.Data.Binary;
+using Color = Microsoft.Xna.Framework.Color;
 
 namespace Ship_Game
 {
@@ -18,6 +19,11 @@ namespace Ship_Game
         [StarData] public string RealDate;
         [StarData] public string ModName = "";
         [StarData] public DateTime Time;
+        // the player's flag, so the save list can show it without a race-name lookup
+        // (a renamed or custom race matches nothing and fell back to the default icon).
+        // -1 = header written before the field existed; the reader falls back to the lookup.
+        [StarData] public int FlagIndex = -1;
+        [StarData] public Color EmpireColor;
     }
 
     public sealed class SavedGame
@@ -84,6 +90,8 @@ namespace Ship_Game
                 RealDate   = now.ToString("M/d/yyyy") + " " + now.ToString("t", CultureInfo.CreateSpecificCulture("en-US").DateTimeFormat),
                 ModName    = GlobalStats.ModName,
                 Time       = now,
+                FlagIndex  = state.Player.data.Traits.FlagIndex,
+                EmpireColor = state.Player.EmpireColor,
             };
 
             // an annoying edge case, someone has created a folder with the same name


gkapulis (approved):
Approved. Clean on every axis, and the backward-compatibility reasoning holds up — I verified it both by reading the serializer and empirically against a real save folder.

**The mechanism behind your `-1` claim, confirmed two levels down:**

- `UserTypeSerializer.CreateInstance()` compiles an `E.New(Type)` expression — a genuine `new T()`, not `FormatterServices.GetUninitializedObject` (`UserTypeSerializer.cs:86-122`). So field initializers do run and `FlagIndex` starts at `-1`.
- `ReadUserClass` drives its loop off `instanceType.Fields`, which comes from the **stream's** type table rather than the current CLR type (`BinarySerializerReader.cs:437-465`). A field absent from an older save is simply never visited, so the `-1` survives.

Worth stating why this mattered: had `CreateInstance` used `GetUninitializedObject`, `FlagIndex` would have landed on `0`, the `>= 0` guard would have passed for every legacy save, and they'd all have drawn flag 0 tinted with a transparent `default(Color)`. The design only works because of that specific construction path.

Rollback direction is safe too — a new save read by an older build hits `fi.Field == null` and skips the field (`BinarySerializerReader.cs:474-475`), the same tolerance that lets us delete `[StarData]` fields.

**Empirical check.** Built this branch and ran `LoadGame.PeekHeader` over a folder of 38 real saves, then wrote one more with this build and re-scanned the same folder:

```
v=21 flag=3  color={R:68 G:140 B:203 A:255}  <- written by this branch
v=21 flag=-1 color={R:0 G:0 B:0 A:0}         <- x37, pre-change saves
RESULT total=38 flagIndexMinus1=37 other=1
```

No exceptions, no version bump, both formats coexisting in one folder, legacy headers all routing to the old name-lookup path. Exactly as described.

**The trap you avoided is worth calling out:** not bumping `SaveGameVersion` was essential. `LoadSaveScreen.InitSaveList` filters on `Version: SavedGame.SaveGameVersion` exact-match, so a bump would have silently hidden every existing save from the load list. The `-1` sentinel is the right mechanism.

Also checked: `Color` has a registered binary serializer (packed uint32); `Traits.FlagIcon` is itself just `ResourceManager.Flag(FlagIndex)` (`RacialTrait.cs:134`) so stock races render identically; and `EmpireColor` is actually *more* correct than the old `Traits.Color` — it's what every other flag draw in the game already uses (`MainDiplomacyScreen.cs:173/210`, `EmpiresButton.cs:157`). The save list was the outlier. An out-of-range index is safe as well (`TryGetTexture` → null → the ctor's Wisp fallback), and can't realistically arise since the list already filters on `header.ModName == GlobalStats.ModName`.

Two things I'll tidy on merge rather than send back:

1. The `using Color = Microsoft.Xna.Framework.Color;` line in `GenericLoadSaveScreen.cs` shows as modified but the text is identical — a trailing-whitespace edit. Reverting it.
2. `EmpireColor` gets no field initializer, so legacy headers read back fully transparent (visible in the scan above). Harmless today since the `FlagIndex >= 0` guard blocks that path, but it's a live trap for anyone who later reads the field without checking. Defaulting it to `Color.White`.

Merging into the `jupiter-1.60/fixes_29` batch. Thanks — the serializer homework on this one was the right call.

gkapulis:
Merged into `jupiter-1.60/fixes_29`. Two corrections to what I said above, plus one real issue found on the deeper pass — none of which change the approval.

**My whitespace nit was wrong.** I claimed the `using Color = ...` line was a trailing-whitespace edit to revert. It isn't: the base file had a lone `LF` on that one line in an otherwise `CRLF` file, and your change normalizes it. That's a cleanup, not noise. Kept as-is.

**`FlagIndex` needed an explicit default attribute.** This is the one worth knowing about, because it's a property of the serializer that isn't visible from the diff. `ObjectScanner.IsDefaultValue` skips any field equal to `owner?.A.DefaultValue ?? ser.DefaultValue` — the `[StarData]` attribute's default, or the TypeSerializer's. **The C# field initializer is never consulted on write.** So with a bare `[StarData] public int FlagIndex = -1`, the writer's skip value is `0`, not `-1` — and flag 0 is a real, pickable flag (`RaceDesignScreen.cs:500-513` wraps the cycler to 0 in both directions).

It round-trips correctly today, but only by accident of `UserTypeState.cs:104-112`: partial layout is chosen only when `validFields < numDefaults`, and `HeaderData`'s 9 mostly-always-set fields keep that false, so full layout always writes an explicit `0` pointer that the reader re-derives back to `0`. Add a handful more optional fields to the header — this PR just added two — and partial layout becomes reachable, `FlagIndex` gets omitted for flag-0 players, and it reads back as the initializer `-1`. That silently restores the exact bug this PR fixes, for flag 0 only. Changed to:

```cs
[StarData(DefaultValue = -1)] public int FlagIndex = -1;
```

which makes `-1` the skipped value (correct — it means "not written") and forces `0` to always be written, independent of layout choice.

Note this also refines the mechanism I described in the approval. Your `-1` reasoning is right, but it's worth being precise about *which* `-1`: for a field **absent from the stream's type table** (a genuine old header) the initializer survives, which is your case and it works. For a field **present with a null pointer**, `BinarySerializerReader.cs:481` explicitly assigns the declared default over the initializer. Two different defaults that happen to coincide here — now they coincide by declaration rather than by luck.

**One narrow behavioral gap.** The new branch commits to the flag path on `FlagIndex >= 0`, but `ResourceManager.Flag()` returns `null` for an index outside the loaded atlas, and `FileData`'s ctor then substitutes `ShipIcons/Wisp` — a ship silhouette in the flag slot. I'd dismissed this initially on the grounds that the list filters by `ModName`, but that's only true of `LoadSaveScreen`; `SaveGameScreen.InitSaveList` lists every parseable save with no mod or version filter, so a modded save with an out-of-atlas index is reachable in the in-game Save dialog. Now falls through to the name lookup when `Flag()` misses, which is a better guess than a wisp.

Also added `SaveHeader_FlagIndexZero_SurvivesRoundTrip` to `SerializationRegressionTests`, covering 0 / -1 / 1 / 7, so the sentinel semantics are pinned rather than depending on a size heuristic staying put.

Re-verified after all of it: 37 pre-change headers still read `-1` and keep the lookup path, one written by this build carries flag + color, no exceptions. Thanks again — the core call here was right, and the failure mode above was well outside what the diff shows.

gkapulis:
Merged into jupiter-1.60/fixes_29 as 690e75fbc (auto-close does not fire for merges off main).

PR #397, opened 2026-08-18
Biospheres become a capacity decision (issue 321 follow-up)

Ludoal:
Follow-up to #321 and the #370 review. Draft, for discussion.

The #370 review's plan (a smoothed AverageTaxRate x the planet's TaxRateMultiplier) kills the oscillation, but the original reporter's deeper point still applies to any tax-based gate, smoothed or not: "The construction of biospheres to increase the population should be independent of the tax rate." Added population contributes production, science and farming with its untaxed share, so judging a biosphere purely on tax revenue systematically undervalues it. In the reporter's own arithmetic: 30% taxes on 20B beats 20% taxes on 10B, because the working (untaxed) population grows from 8B to 14B.

This draft takes that point to its conclusion: the tax rate leaves the biosphere decision entirely.

- **BUILD** when the colony runs out of room: population at 85%+ of the cap, or a build list with no free habitable tile (the blueprint-completion case from the original report).
- **SCRAP** a free biosphere only when the population would sit under 60% of the cap without it. The wide 60/85 dead band makes build/scrap oscillation impossible by construction - no smoothing needed.
- Money keeps exactly one word: the colony budget (can the upkeep be afforded), with a "budget covers it, keep the last one" guarantee for player-funded biospheres.

Deliberately narrow: `BioSphereProfitable` is removed with its criterion; `NetRevenueGain` and the tax machinery (racial TaxMod, PlusTaxPercentage buildings) are untouched for every other building, so the concerns from the #370 review do not arise here.

Test status: compiled and logic-reviewed; running on our fork's builds. Biosphere governor behavior only shows on long campaigns, which is exactly why this is a draft - the original reporter has the worked reproduction scenarios, and we'd value a pass with them before this leaves draft.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/ColonyResource.cs
+++ b/Ship_Game/Universe/SolarBodies/ColonyResource.cs
@@ -273,6 +273,10 @@ namespace Ship_Game.Universe.SolarBodies
         // The current tax rate applied by empire tax rate and planet tax rate modifiers
         public float TaxRate { get; private set; }
 
+        // the STRUCTURAL part of the tax pipe: racial bonus/penalty plus building tax
+        // percentages - everything but the player's slider (PR 397 review)
+        public float TaxRateMultiplier { get; private set; } = 1f;
+
         // revenue before maintenance is deducted
         public float GrossRevenue { get; private set; }
 
@@ -324,6 +328,7 @@ namespace Ship_Game.Universe.SolarBodies
             TroopMaint = Planet.Troops.Count * ShipMaintenance.TroopMaint; // We count enemy troops as well
 
             // And finally we adjust local TaxRate by the bonus multiplier
+            TaxRateMultiplier = taxRateMultiplier;
             TaxRate     *= taxRateMultiplier;
             Maintenance *= Planet.Owner.data.Traits.MaintMultiplier;
 
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet.cs
@@ -1521,19 +1521,36 @@ namespace Ship_Game
             else if (!RecentCombat)
             {
                 // population is increased
+                Population += EstimatedPopGrowthPerTurn;
+                Population  = Population.Clamped(0, MaxPopulation);
+            }
+
+            Population = Math.Max(10, Population); // over population will decrease in time, so this is not clamped to max pop
+        }
+
+        // The per-turn population gain, extracted from GrowPopulation so callers can ask
+        // whether (and how fast) a colony is actually growing. Same arithmetic the tick
+        // applies (rep rate with the trait clamps, flat bonus and reproduction mod, then the
+        // short-on-food factor), so the number matches what the colony grows by. Zero when
+        // the colony is not in the growing branch (starving, over-populated, recent combat).
+        public float EstimatedPopGrowthPerTurn
+        {
+            get
+            {
+                if (Owner == null || RecentCombat || !CanRepairOrHeal()
+                    || PopulationRatio.Greater(1) || IsStarving)
+                    return 0f;
+
                 float balanceGrowth = (1 - PopulationRatio).Clamped(0.25f, 1f);
                 float repRate       = Owner.data.BaseReproductiveRate * (Population/3) * balanceGrowth;
                 if (Owner.data.Traits.PopGrowthMax.NotZero())
                     repRate = repRate.UpperBound(Owner.data.Traits.PopGrowthMax * 1000f);
 
-                repRate     = repRate.LowerBound(Owner.data.Traits.PopGrowthMin * 1000f);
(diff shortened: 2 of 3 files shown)


Ludoal:
Updated per the review: biospheres now come in the two suggested classes.

- **Facilitation**: a building the colony wants has no free tile. Judged against the colony budget on the combined upkeep (biosphere plus the building it unlocks). Desire-based rather than queue-based, since a building cannot even be enqueued without a free tile.
- **Population capacity**: population pressure against the cap, filtered by the economics suggested in the review. The added population must pay the biosphere's upkeep at a reasonable rate: upkeep <= 0.6 x the pop's full-rate marginal revenue on that planet (0.6 is a named constant, mid of the suggested 50-75% band).

The tax slider still plays no part in the decision, and the 60/85 dead band still rules out build/scrap oscillation. The small `EstimatedPopGrowthPerTurn` extraction from `GrowPopulation` (same arithmetic, now callable) supports the growth check.

Ludoal:
Good catch on the racial tax modifiers. Done: the marginal income is now computed directly as (pop the biosphere adds) x (income per colonist) x (structural tax multiplier). The structural multiplier is the racial bonus/penalty plus the building tax percentages, exposed by `ColonyMoney.Update` (the same value it already folds into the live tax rate). The player's tax slider remains excluded from the decision.

gkapulis:
Thank you for this, and for the two rounds of revision. The capacity model is the right answer and it has landed in-house as **933196b4f** on `jupiter-1.60/fixes_33`, adapted from this draft. Closing here rather than merging, because the review turned up two things that needed changing and it was easier to carry them along with the surrounding cleanup than to ask you for a third revision.

**What carried over unchanged:** the two classes of biosphere, the facilitation case judged on the combined upkeep, the 85/60 dead band, the payback share at 0.6 as the middle of the reporter's 50-75% band, and `TaxRateMultiplier` exposed from `ColonyMoney.Update` so the payback test sees the structural modifiers and never the player's slider. That shape is yours and it survived intact.

**What changed, and why:**

1. **The added population is `PopPerBiosphere(Owner)` alone.** `bio.MaxPopIncrease` contributes nothing to a planet's cap: `UpdateMaxPopulation` builds `PopulationBonus` as `SumBuildings(b => !b.IsBiospheres ? b.MaxPopIncrease : 0)`, and biospheres enter only through `MaxPopValFromTiles`. The in-game description says the same thing ("max population based on half of the Colonists per Tile"). Vanilla `PopPerTile` runs 40-400, so counting the template's 100 overstated the income term by 2x to 9x, worst on the hostile low-environment worlds where biospheres matter most. That would have let the payback test pass on almost anything, which is the one thing the reporter's 50-75% band exists to prevent. The dead field is on our list to delete from `Biospheres.xml`.

2. **Biospheres had to be filtered out of the wanted list.** `GetBuildingsWeCanBuildHere` drops the biosphere only when the whole planet is already habitable, which is the case `TryBuildBiospheres` has already returned on. So `wanted` always contained a biosphere: it could be its own cheapest "wanted building" for the facilitation budget test, and `wanted.Count > 0` was permanently true, which meant the last free biosphere could never be scrapped. One `Filter(b => !b.IsBiospheres)` fixes both.

3. **The `GrowPopulation` extraction is out.** This one is worth knowing about for future work in that file: `CanRepairOrHeal()` reads like a predicate but is `BombingIntensity == 0 || Random.RollDice(100 - BombingIntensity)`. Copying it into a property that the governor also calls meant a bombed planet rolled twice per turn (p squared instead of p), and the AI path drew extra values from the sim RNG stream. We dropped the growth gate instead, which removed the need for the property at all: `balanceGrowth` floors at 0.25, so growth is positive at any population ratio at or under 1, and the gate could only ever fire under recent combat. Losing it also fixed a side effect nobody had spotted - an over-populated colony has zero growth, so a growth gate refuses a biosphere exactly where the roof is most needed.

4. Smaller ones: the three magic numbers became named constants so mods can find them; a guard keeps the last free biosphere while the colony still has something to build, which protects your original blueprint-completion report; and `HabiableBuiltCoverage` was deleted after losing both its readers.

A separate commit (**a5c45f34b**) stops the colony screen tinting biospheres "unprofitable", which was still the old tax rule and would have contradicted the governor on screen. That removed `NetRevenueGain` entirely - it had no callers left, and it is where the `MaxPopIncrease + PopPerBiosphere` arithmetic originally came from, so the double count is gone at its source.

Eight unit tests cover the new rules, each verified by re-introducing the bug and watching the matching test fail. One of them sets the tax rate to 0.1 and then to 1.0 and asserts the decision does not move, so #321 cannot quietly come back.

Your diagnosis work has been consistently good and this was no exception - the design was right, and the reporter's specification came through it faithfully. Thanks again.

gkapulis:
Superseded by 933196b4f on jupiter-1.60/fixes_33; see the previous comment for what carried over and what changed.

PR #398, opened 2026-08-19
MiniMap: the viewport frame stops trembling far from the galaxy centre

Ludoal:
## Symptom

The minimap viewport frame trembles for a few moments while the camera eases, with an amplitude proportional to the camera's distance from the galaxy centre (maintainer bench report).

## Cause

The frame is rebuilt every draw from float `UnprojectToWorldPosition` round-trips. Float precision degrades as world coordinates grow, so the input itself is noisy far from the origin, and no display-side rounding or damping can fix a noisy source (the existing `RoundTo(x, 1)` calls were placed for exactly this and cannot recover the lost precision).

## Fix

Frustum geometry is linear in camera height, so the frame now derives from camera state instead: one calibration against the real projection captures width-per-height and centre-offset-per-height, and every later frame is pure arithmetic on the double-precision camera position. One pair of Unproject calls per frame goes away with it.

Benched on the fork at every zoom level, camera distance and easing speed: the tremble is gone.


gkapulis:
Landed as **8b182aa2b** on `jupiter-1.60/fixes_33`, with tests. Closing here; the implementation differs enough that it was cleaner to apply in house than to ask for another round.

**Your diagnosis is right, and it is the more interesting kind of right.** This is not a minimap bug — it is our known `Matrix.Invert` precision problem surfacing somewhere visible. `UnprojectToWorldPosition` inverts the view-projection, `CreateLookAtDown` casts the camera position to float, and the error grows with the magnitude of world coordinates. `RoundTo(x, 1)` quantised at one world unit, which is far below that noise floor, so it damped nothing at all. Every part of that chain checks out.

**Where the implementation goes wrong is subtle and worth spelling out**, because the idea of calibrating is sound and it is only the source of the calibration that is not.

`FrustumOffsetXPerHeight` and `FrustumOffsetYPerHeight` are measured as `(frustum centre - CamPos) / camH`, where `frustum` is `VisibleWorldRect` — which is itself `UnprojectToWorldRect`, so it carries exactly the error you are escaping. For a camera that looks straight down, the true value of both offsets is **zero**. So the calibration can only ever capture one sample of the noise, and the code then multiplies that sample back by camera height on every later frame at every position. The tremble becomes a permanent, height-proportional bias — wrong even at the galaxy centre, where the old code was right. `FrustumWidthPerHeight` has the same origin.

The one-shot latch has a second problem: nothing invalidates it. If the viewport aspect changes the constants are stale, and if the first eligible frame is degenerate the frame stays clamped at 2x2 forever. The old code was at least self-healing each frame.

**None of it is necessary, because the camera does not need inverting at all.** `CreateLookAtDown` always aims from `(camX, camY, -camZ)` at `(camX, camY, 0)`, so the visible rect at the z=0 plane is centred on `CamPos` exactly — and `CreatePerspectiveFieldOfView` puts the rest in the projection, where `M11` is `1/(tan(fov/2)*aspect)` and `M22` is `1/tan(fov/2)`. So the half extents are `camZ/M11` and `camZ/M22`. That is now `UniverseScreen.ExactVisibleWorldRect`: exact at any distance, stateless, self-correcting on a resolution change, and two divisions where there were two matrix inversions.

It also closes a second jump your version would have kept. The old height came from the width times the *screen* aspect while the width came from the *projection's* aspect, and the projection is rebuilt per `viewState` with a different far plane. Near and far land in `M33` and `M43`, so reading `M11` and `M22` is immune to that. There is a test for it.

**I measured the tremble rather than assuming the fix removed it**, since the symptom is visual and the mechanism is not — and my first attempt at a test was useless, because it derived the frame position from `CamPos` directly and so could not tell the two implementations apart. The version that works pans the camera 2000 units a frame from nine million out: the old frame **reverses direction 33 times over 300 frames while advancing 26 pixels**, and the new one reverses none. That is the reported tremble, and it is now a test.

Two small things: the branch also committed `game/Content/Textures/.DS_Store`, so `.gitignore` has an entry for it now; and the explanatory comments came out, since reasoning goes in the commit message here.

One thing I did not do, and it is the real follow-up if you want it: `GameScreen.VisibleWorldRect` still inverts, and every other consumer of it — frustum culling, the object manager, the influence tree — still carries the same drift. They are cull rects, so it shows as edge popping rather than tremble, which is why nobody has reported it.

Thanks for this one. Tracing a visible wobble back to the matrix inversion behind it is not an obvious step.

PR #399, opened 2026-08-25
Serializer: skip enum values from deleted types instead of crashing on load

Ludoal:
Loading a save that carries an enum type the running build does not know crashes with a NullReferenceException in `BinarySerializerReader.ReadObjects`.

The header pass already handles the case: `GetTypeFrom` returns null, `AddDeletedTypeInfo` records the type with `Ser = null` and logs "Read DELETED". The object pass then calls `ser.Deserialize(this)` on that null serializer.

Enum values are a single VLi32 each, so they can be read past without knowing the type. This adds that branch, guarded on `ser == null` and the Enums category, so nothing else changes.

The other deleted categories still throw. Skipping those needs the `CollectionSkippingSerializer` the FIXMEs at lines 219 and 398 describe, which is a larger change. Happy to submit it separately if there is interest.

Tested in a live build: the game starts and existing saves load normally. The crash path itself needs a build without the enum reading a save with it, which is the situation this fixes rather than one I can reproduce in a single build.

### DIFF
--- a/Ship_Game/Data/Binary/BinarySerializerReader.cs
+++ b/Ship_Game/Data/Binary/BinarySerializerReader.cs
@@ -399,6 +399,18 @@ public class BinarySerializerReader
                 cs.Deserialize(this, instance);
             }
         }
+        // an enum type that no longer exists: its values are still in the stream and each one
+        // is a single VLi32, so they can be read past without knowing the type. Every other
+        // deleted category still needs a skipping serializer (see the FIXMEs above).
+        else if (ser == null && type.Category is SerializerCategory.Enums)
+        {
+            if (Verbose) Log.Warning($"Skipping {count} values of deleted enum {type}");
+            for (int i = 0; i < count; ++i)
+            {
+                BR.ReadVLi32();
+                ObjectsList[baseId + i] = null;
+            }
+        }
         // fundamental types such as int, Vector2, String, Byte[], Point. @see TypeSerializerMap.cs
         // also RawArrays like Ship[]
         else


gkapulis:
Landed as **5a79efef1** on `jupiter-1.60/fixes_33`, with a test. Closing here since the branch now carries it.

The analysis in the PR body is exactly right, and I checked the two things it rests on rather than taking them on trust, because a wrong skip width in a binary stream is worse than the crash it replaces — it would silently corrupt every group after it.

- `EnumSerializer.Serialize` is unconditionally one `WriteVLi32` and nothing else. The several `GetOrdinal*` paths only choose how the ordinal is *obtained*, never the width written, and an underlying type that will not fit an int throws in the `EnumSerializer` constructor, so such an enum can never reach a stream in the first place. There is no per-object framing around it either: enums get the plain `ObjectState`, and `WriteGroup` emits only typeId and count before the loop, both of which `ReadObjects` has already consumed.
- The `Enums` category comes from the stream's own flag bit, not from the resolved type. That is what makes the branch reachable at all for a type that no longer exists.

I dropped the `ObjectsList[baseId + i] = null;` from your version. `PreInstantiate` goes through `GetTypeGroups`, which filters on `Type?.Ser != null`, so a deleted group is never instantiated and those slots are already null.

**On testability** — you wrote that the crash path needs a build without the enum reading a save with it, which you could not do in one build. You can, and the file you were working next to already does it: `CreateByteStreamForDeletedTypeTest` plus the commented-out type is exactly that pattern, used by `ContainsDeletedTypes` and `ContainsMovedTypes`. Generate the stream with the type present, comment the type out, keep the base64. `ContainsDeletedEnums` does that now.

One detail from writing it that is worth passing on. My first version used `Two` and `Three` with their default ordinals, 2 and 3. Both encode to a single VLi32 byte — so a skip written as `ReadByte()` per value consumes exactly the same bytes and the test passes green, which makes it useless against the one mistake that actually matters here. The ordinals are now `Two = 4242` and `Three = -7`: 4242 takes two bytes and -7 sets the sign bit. Verified both ways — the load NREs with the fix removed, and throws `InvalidDataException` with `ReadByte` in place of `ReadVLi32`, which the weak version did not catch.

Your scope note is in the commit message, because it is worth not over-claiming: this covers scalar enum fields. An enum inside a collection still fails the load at the "Missing StreamType" throw, and `Map<DeletedEnum, T>` dereferences the deleted key's type at `AddCollectionTypeInfo` before it ever reaches `ReadObjects` — the guard there covers only the value type, though the error-message path two lines below does handle a null key. Enum-keyed maps are live in real saves (`EmpireData`'s `Map<WeaponTag, WeaponTagModifier>`, `Remnants`' `Map<RemnantShipType, float>`), so that one is worth its own small fix.

**Yes to the separate PR for the collection-skipping serializer**, if you still want to take it on — it is the FIXME at both sites and it would close the rest of this class. Worth saying it is the harder half: a collection's element count and per-element framing are not uniform the way a VLi32 is.

PR #401, opened 2026-09-01
Blueprints: the terraformer half-completion gate never closed

Ludoal:
`ColonyBlueprints.IsHalfAchievableCompleted` compared the reachable share of a plan against half of the *completed* share:

```csharp
PercentAchievable >= PercentCompleted/2
```

A standing building counts as reachable too, so `PercentAchievable` is never below `PercentCompleted`. The test therefore read true in every case, and on any plan claiming at least half the tile area — the branch of `OkToBuildTerraformers` that uses this gate — the gate never closed. Terraformers could go up on a colony that had barely started its list.

It now asks what the name says: half of what this colony can reach is already up.

```csharp
PercentCompleted >= PercentAchievable/2
```

One line, plus a comment recording why achievable cannot sit below completed.

Reported independently by a player who noticed terraformers appearing early on freshly colonised worlds while their plans were still running, without knowing the cause.

### DIFF
--- a/Ship_Game/Universe/SolarBodies/ColonyBlueprints.cs
+++ b/Ship_Game/Universe/SolarBodies/ColonyBlueprints.cs
@@ -27,7 +27,9 @@ namespace Ship_Game.Universe.SolarBodies
         bool Completed => PercentCompleted == 100;
 
         public bool IsAchievableCompleted => PercentAchievable == 0 || PercentAchievable == PercentCompleted;
-        bool IsHalfAchievableCompleted => PercentAchievable == 0 || PercentAchievable >= PercentCompleted/2;
+        // a standing building counts as reachable too, so achievable is never below completed:
+        // the gate asks whether half of what this colony can reach is already up.
+        bool IsHalfAchievableCompleted => PercentAchievable == 0 || PercentCompleted >= PercentAchievable/2;
 
         public bool IsRequired(Building b) => PlannedBuildings.Contains(b.Name);
         public bool IsNotRequired(Building b) => !IsRequired(b);


gkapulis:
Applied directly as **95931bf65** on `jupiter-1.60/fixes_33`, with tests. Closing this rather than merging only because the fix went in alongside a follow-up it uncovered; the one-line change is yours as written.

The diagnosis holds up completely. Both percentages divide by the same `PlannedBuildings.Count`, so the comparison reduces to buildings standing against buildings unlocked, and since anything standing on your own colony was unlocked in order to build it, achievable is never below completed. `PercentAchievable >= PercentCompleted/2` was therefore true whatever the colony had built, and the gate was dead for every large plan. `PercentCompleted >= PercentAchievable/2` is what the name says.

One nuance on the write-up, for the record rather than as a correction: it is not quite true in *every* case. `PercentCompleted` counts building instances while `PlannedBuildings` is a set of names, so a non-unique planned building could in principle be counted more than once and push completed above achievable. In practice it cannot happen through the UI — `IsSuitableForBlueprints` excludes Biospheres and the Terraformer, the only two non-unique buildings shipped — so your reasoning holds everywhere it matters. It does point at a gap in `ResourceManager.BlueprintsValid`, which checks only that each planned building exists; we have logged that separately.

Three tests now cover the gate: a barely started large plan holds terraformers back, a plan past halfway lets them in, and a plan the empire cannot build at all never blocks. Each was checked by restoring the original expression and confirming the matching test fails.

**What closing the gate uncovered**, which is worth knowing because it was unreachable while the gate was dead. `Planet.UpdateTerraformBudget` is the only consumer, and `Empire.UpdateTerraformerBudget` funnels through it, so this reaches the empire's automatic terraforming as well as the governor. Under an *exclusive* plan, `PlayerBuiltIsProtected` is `OwnerIsPlayer && !HasExclusiveBlueprints`, so the player-added guard in `TryCancelOverBudgetCivilianBuilding` does not fire — and the next branch cancels any queued terraformer while `TerraformBudget == 0`, which a closed gate guarantees. The player could not keep a hand-queued terraformer on such a colony at all, with no escape but turning AutoBuildTerraformers off empire-wide, on exactly the worlds where terraforming is what lets the plan finish.

The maintainer's call is that terraformers are outside the plan's jurisdiction: exclusivity clears competing buildings, but a terraformer is what makes the planet able to hold the plan. The code already agreed in two places — `SuitableForScrap` skips anything with `PlusTerraformPoints`, and `IsSuitableForBlueprints` excludes terraformers, so a standing one was always safe. Only the queue disagreed. Fixed in **300dc9ad3**: that cancel now leaves a player-queued terraformer alone, while the governor's own are still cancelled when it cannot pay.

Thanks for this one. A dormant gate that silently never fires is a hard class of bug to notice, and the reasoning that found it was exactly right.

gkapulis:
Applied directly as 95931bf65 with tests, plus the follow-up it uncovered in 300dc9ad3. See the previous comment.

PR #402, opened 2026-09-01
Refit: prioritized ports should narrow the choice, not replace the selector

Ludoal:
`FindPlanetToRefitAt` handed the entire choice to `FindPlanetToBuildShipAt` when the player has prioritized ports:

```csharp
if (isPlayer && PlayerPrioritizedPorts.Length > 0)
    return FindPlanetToBuildShipAt(ports, newShip, out planet);
```

That selector ranks candidates by queue time alone, which is right for a hull that does not exist yet. A refit is different: an existing ship has to fly to the port, and back if `travelBack` — which is why the method's own selector, two lines below, already includes `GetAstrogateTimeTo`. Taking the other branch dropped that term, so a ship could be sent across the map to a marginally emptier yard.

Prioritized ports now restrict the candidate list without replacing the selector:

```csharp
IReadOnlyList<Planet> actualPorts = isPlayer && PlayerPrioritizedPorts.Length > 0 ? PlayerPrioritizedPorts : ports;
```

The player's preference is still honoured — it just narrows the field instead of changing how the field is ranked.

### DIFF
--- a/Ship_Game/Empire_RallyPlanets.cs
+++ b/Ship_Game/Empire_RallyPlanets.cs
@@ -218,10 +218,11 @@ public sealed partial class Empire
         planet = null;
         int travelMultiplier = travelBack ? 2 : 1;
 
-        if (isPlayer && PlayerPrioritizedPorts.Length > 0)
-            return FindPlanetToBuildShipAt(ports, newShip, out planet);
+        // Prioritized ports restrict the candidates, but they must not change the selector:
+        // an existing ship has to fly there (and back), so travel time stays part of the choice.
+        IReadOnlyList<Planet> actualPorts = isPlayer && PlayerPrioritizedPorts.Length > 0 ? PlayerPrioritizedPorts : ports;
 
-        if (ports.Count == 0 || !GetBestPorts(ports, out Planet[] bestPorts, 1))
+        if (actualPorts.Count == 0 || !GetBestPorts(actualPorts, out Planet[] bestPorts, 1))
             return false;
 
         planet = bestPorts.FindMin(p => p.TurnsUntilQueueComplete(cost, 1f, newShip)


gkapulis:
Landed as **3b01a5bc7** on `jupiter-1.60/fixes_33`, with tests. Closing rather than merging, because our commit now touches the same lines and would collide — but the diagnosis here was right, and it found a bug I wrote.

`git blame` puts those two delegation lines at **7d034cbee** (Mars 1.50, PR #182), so this was mine. It is player-only, since `PlayerPrioritizedPorts` is populated under `isPlayer`, and with a single prioritized port both paths agree — it takes two before anything diverges, which is presumably why it sat there for a year.

**The astrogate term is the part a player would actually feel, and you found it.** I checked the units before trusting the restored selector, since a sum of incommensurable terms would be worse than either half: `GetAstrogateTime` divides by `Universe.P.TurnTimer`, so it really is in turns, same as `TurnsUntilQueueComplete`. The sum is sound.

Two more things came off with the delegation that the PR does not mention, and they push the same way:

- **The cost.** `FindPlanetToBuildShipAt` takes no cost parameter — it recomputes `ship.GetCost(this)` and silently discards the `cost` argument the caller passed. Callers pass `OldShip.RefitCost(...)`, which for a same-hull refit is roughly a tenth of the full price. That does not scale away harmlessly: in `TurnsUntilQueueComplete` an inflated cost weighs raw production against existing queue depth differently, so the ranking itself moves, not just the magnitudes.
- **portQuality**, 1 in the refit selector against the build default of 0.5.

Not delegating removes all three at once, which is what your change does.

Two things we added on top:

- **The second overload** at `Empire_RallyPlanets.cs:230` had the identical delegation, fifteen lines down — same commit of mine, used by `RefitOrbital`. No astrogate term applies there (the old station never flies; a constructor is built at the chosen port and deploys out to it), but the cost and portQuality drift are the same. Fixing one and leaving the other seemed worse than fixing neither.
- That overload also had an early `ports.Count == 0` return sitting *before* the prioritized check, so a player whose safe ports had all gone unsafe got no refit at all despite having named a yard. It now matches the ship overload, which never had that return.

The two-line comment came out — reasoning goes in the commit message in this repo.

One thing worth stating plainly rather than leaving implied: the PR body describes this as restricting the candidate list, but `actualPorts` *replaces* the caller's list rather than intersecting it, so a prioritized port in a contested system can still be chosen for a ship that ignores combat on the way there. That is pre-existing and shared with every build goal, so it is not this change's job — but it is not quite "narrowing" either, and I would rather say so than let it read as fixed.

Tests are in `UnitTests/Empires/RefitPortSelectionTests.cs`. Five, of which two fail with the fix reverted — verified, not assumed. The main one carries a control assertion first: the travel-blind overload picking the *far* port, which passes under the revert, so the failure is the fix doing its job and not a fixture that quietly stopped discriminating.

The in-game Codex needed correcting alongside, and it was interesting to see how it had drifted: the Refitting entry put "counting the flight there, and back again for a ship from a fleet" inside the *otherwise* branch, which documented the bug perfectly faithfully. Both entries now cover the prioritized branch, with orbital refits called out separately since nothing flies on that path.

Thanks for this one — it is the kind that produces no error, no log line and no crash, just ships quietly taking the long way around for a year.

PR #403, opened 2026-09-01
Blueprints: a technology unlock refreshed only half the pair

Ludoal:
`Empire.UpdateBlueprintsAchievablePercentage` recomputes only one of the two figures a blueprint carries:

```csharp
p.Blueprints?.UpdatePercentAchievable();
```

`PercentCompleted` and `PercentAchievable` share a denominator — the plan's entry count — and a technology unlock changes what a colony can reach. Recomputing the achievable share alone leaves the pair disagreeing until something else happens to refresh them, and the colony screen shows a plan further along, or further behind, than it is.

Both are now recomputed together, using the two methods `ColonyBlueprints` already exposes:

```csharp
p.Blueprints?.UpdateCompletion();
p.Blueprints?.UpdatePercentAchievable();
```

Note on naming: the method now updates both figures while still being called `UpdateBlueprintsAchievablePercentage`. Left as is to keep the diff to the fix — happy to rename it in this PR if you would rather.

### DIFF
--- a/Ship_Game/Empire.cs
+++ b/Ship_Game/Empire.cs
@@ -1005,6 +1005,10 @@ namespace Ship_Game
             for (int i = 0; i < OwnedPlanets.Count; i++)
             {
                 Planet p = OwnedPlanets[i];
+                // the whole pair: a technology changes what a colony can reach, and the figure
+                // beside it shares that denominator, so recomputing one alone leaves the two
+                // disagreeing until something else refreshes them.
+                p.Blueprints?.UpdateCompletion();
                 p.Blueprints?.UpdatePercentAchievable();
             }
         }


gkapulis (changes requested):
I don't think this one holds up — the premise doesn't survive reading the two methods, and the change is not a harmless no-op.

## The two figures don't move together

The argument is that they "share a denominator... so recomputing one alone leaves the pair disagreeing." A shared denominator only matters if the denominator moves. Here is what each figure actually reads (`ColonyBlueprints.cs:68-84`):

| | numerator | denominator |
|---|---|---|
| `PercentCompleted` | `P.Buildings.Count(IsRequired)` — buildings physically built | `Template.PlannedBuildings.Count` |
| `PercentAchievable` | `unlockedBuildings.Count(IsRequired)` | `Template.PlannedBuildings.Count` |

A technology unlock does not construct or demolish a building, and it does not mutate the template. So **neither input to `PercentCompleted` changes**, and it cannot be stale at that point. The only thing a tech unlock moves is `PercentAchievable`'s own numerator — which is exactly what the existing single call recomputes.

What is already there is a correct set of event-driven updates, each fired precisely when its own inputs change:

- `PercentCompleted` ← `UpdatePlanetStatsFromPlacedBuilding` (`Planet_ConstructionQueue.cs:446`) and `UpdatePlanetStatsFromRemovedBuilding` (`:476`) — the only two paths that mutate `BuildingList`, since `PlaceBuildingAt` is documented as "the only way to add a building to the planet!"
- `PercentAchievable` ← `Empire.UpdateBlueprintsAchievablePercentage`, the path this PR touches
- Both ← `OnTemplateChanged`, i.e. when the shared denominator genuinely changes

So the added call recomputes a value from inputs that have not changed.

## It also carries a side effect

`UpdateCompletion` does not only assign a percentage. It ends with:

```csharp
if (Completed)
    ChangeTemplateIfLinked();
```

which runs `ChangeTemplate` → `OnTemplateChanged` → `ChangeColonyType()`, and that writes `P.CType`.

Blueprints and their `LinkTo` chains are player-authored through `BlueprintsScreen`, so a completed linked template is ordinary player data, not an edge case. With this change, every tech unlock re-runs the template switch on every owned planet whose blueprint reads 100% — overwriting the colony type, including one the player set by hand in the governor dropdown (`GovernorDetailsComponent.cs:395`). Today that is only reachable when a building is added or removed; this makes it fire on every unlock as well.

It also re-enters `OnTemplateChanged` → `UpdateCompletion` → potentially `ChangeTemplateIfLinked` again. `BlueprintsTemplate` already carries explicit self-link guarding, which suggests loops in this data are a known concern.

Cost is secondary but not nothing: `P.Buildings.ToArray()` allocates a copy per planet per unlock, plus a LINQ `Count` with a `HashSet` predicate — dozens of throwaway arrays per unlock on a large empire, to recompute a figure that cannot have moved.

## What did you actually see?

You describe a concrete symptom — the colony screen showing a plan further along, or further behind, than it is. I would rather chase that than close it with this, because the cause has to be somewhere else.

One candidate worth checking: `ColonyBlueprints` captures its empire once, as `[StarData] readonly Empire Owner`, at construction (`Planet_Govern.cs:266`). `UpdatePercentAchievable` then counts against `Owner.GetUnlockedBuildings()`. If a conquered planet keeps its existing `Blueprints` instance, that figure would be computed against the *previous* owner's unlocked buildings, and would stay wrong until the template changed. I have not confirmed whether conquest clears or rebuilds `Blueprints` — flagging it as the first place I would look, not as a diagnosis.

If you can say what you were doing when the numbers looked wrong (conquest, a template switch, a governor change, loading a save), that should narrow it quickly.

On the naming question in the description: no rename needed — if the method keeps updating only the achievable figure, the current name is accurate.

Ludoal:
You're right: the premise doesn't hold. A tech unlock doesn't change what those percentages count, and our refresh was overwriting a player-set colony type in the process. Closing this.

Your note about the owner reference surviving a conquest is a good lead; we'll chase the original symptom from there. Thanks for the close read.

PR #404, opened 2026-09-07
Terraformer picks a terraformable tile deterministically

Ludoal:
`PickTileForTerraformer` returns `Random.Item` over the tiles that pass the volcano filter, so the terraformer lands anywhere on the grid.

Two consequences:

* Since #337, biospheres skip terraformable tiles once terraformers are unlocked, precisely to keep those tiles for the terraformer. The random pick gives that up on the other side: the terraformer often settles on a plain tile and leaves the terraformable one for the next biosphere to cover over.
* Reloading the same save puts the terraformer on a different tile every time, which makes any report about colony layout impossible to reproduce.

The pick becomes: the first terraformable tile in grid order, and when there is none, the last eligible tile, which is the far end from where a biosphere starts looking. The volcano filter and both call sites are unchanged.

Tested on a save where a terraformer kept moving around on a crowded colony.

Companion to #405, which applies the same rule to a biosphere placed by hand. The two are independent, neither depends on the other.

https://claude.ai/code/session_01XiWUe8Cn9QRu181S8oheXy

### DIFF
--- a/Ship_Game/Universe/SolarBodies/Planet/Planet_EvaluateBuildings.cs
+++ b/Ship_Game/Universe/SolarBodies/Planet/Planet_EvaluateBuildings.cs
@@ -744,7 +744,13 @@ namespace Ship_Game
                     potentialTiles.Add(tile);
             }
 
-            return Random.Item(potentialTiles.Count > 0 ? potentialTiles : tileList.ToArrayList());
+            // A terraformable tile first: biospheres already skip those once terraformers are
+            // unlocked, so this is the terraformer's own ground. The LAST of the others when there
+            // is none, which is the far end from where a biosphere starts looking. Deterministic
+            // either way, so reloading a save gives back the same build order.
+            Array<PlanetGridSquare> eligible = potentialTiles.Count > 0 ? potentialTiles : tileList.ToArrayList();
+            PlanetGridSquare terraformable = eligible.Find(t => t.Terraformable);
+            return terraformable ?? eligible[eligible.Count - 1];
 
             bool NoVolcanosAround(PlanetGridSquare tile)
             {


gkapulis:
Applied in-house as **54e182a1f** on `jupiter-1.60/fixes_33`, with tests. This one was mergeable as written — the logic is yours unchanged, and it went in alongside #377 only because both picks live in the same file and wanted the same pass.

Both halves of the reasoning check out. `Random.Item` meant the terraformer competed with biospheres for the non-terraformable ground that biospheres specifically need, and it meant reloading a save laid the colony out differently every time, which makes any report about colony layout impossible to reproduce.

I did want to be sure a terraformer standing on terraformable ground is not stealing from itself, since that ground is exactly what it is there to turn. It is not: `CompleteTileTerraforming` filters `!t.Habitable && t.Terraformable || t.BioCanTerraform` without excluding tiles that carry a building, and `MakeTileHabitable` only destroys one when `LavaHere`. So the tile is still terraformed on schedule, it just has a terraformer standing on it while that happens, and `RemoveTerraformers` frees it when the work finishes. The one consequence worth knowing is that the freshly habitable tile adds nothing to `FreeHabitableTiles` until then.

Both callers guard `Length > 0`, so `eligible[eligible.Count - 1]` cannot go out of range, and `Array<T>.Find` returns null for no match so the `??` behaves.

The only edit was dropping the explanatory comment block — house style keeps that reasoning in the commit message, and it is recorded there, including why the fallback takes the last eligible tile rather than the first: it is the far end from where the biosphere pick starts looking, so the two collide as little as possible.

Two tests cover it: the terraformer takes terraformable ground when there is any, and the same input gives the same tile across repeated calls. The second one fails against `Random.Item`, which is what pins the determinism.

Thanks — a clean diagnosis, and the reproducibility argument is the kind of thing that pays off every time someone files a colony-layout bug.

gkapulis:
Applied in-house as 54e182a1f with tests; the logic is unchanged, see the previous comment.

PR #405, opened 2026-09-07
Biospheres placed by hand follow the governor's tile rule

Ludoal:
Double-clicking a biosphere in the Colony screen drops it on a random free tile: `AssignBuildingToTile` picks with `Random.Item`, and nothing on that path knows about terraformable ground.

The governor already knows. `GetPreferredTile` in `Planet_EvaluateBuildings` takes the first tile that is neither habitable nor terraformable once terraformers are unlocked, and only falls back to a terraformable one when the colony has nothing else free. The manual path never got that rule, so the same colony behaves in two different ways depending on who places the building, and a hand placed biosphere can take the tile the terraformer was waiting for.

`AssignBuildingToTile` now applies the governor's rule to biospheres: leave the terraformable tiles alone while there is something else free, and take the first free tile in list order instead of a random one, so the same colony is laid out the same way twice. A world made of nothing but terraformable tiles still gets its biosphere, and the rule only applies once the terraformer is unlocked, since before that there is nothing to spare the tiles for. Every other building keeps the random pick.

Companion to #404, which fixes the terraformer's own tile pick. The two are independent, neither depends on the other.

https://claude.ai/code/session_01XiWUe8Cn9QRu181S8oheXy

### DIFF
--- a/Ship_Game/Building.cs
+++ b/Ship_Game/Building.cs
@@ -416,8 +416,25 @@ namespace Ship_Game
                 return where.CanEnqueueBuildingHere(b);
 
             PlanetGridSquare[] freeSpots = planet.TilesList.Filter(pgs => pgs.CanEnqueueBuildingHere(b));
+
+            // A Biosphere leaves the terraformable tiles to the Terraformer, and takes the first
+            // free tile in list order instead of a random one, so the same colony is laid out the
+            // same way twice. Only when there is something else to take: a world made of nothing
+            // but terraformable tiles still gets its Biosphere.
+            if (b.IsBiospheres && planet.Owner?.IsBuildingUnlocked(TerraformerId) == true)
+            {
+                foreach (PlanetGridSquare t in freeSpots)
+                {
+                    if (!t.Terraformable)
+                    {
+                        where = t;
+                        return true;
+                    }
+                }
+            }
+
             if (freeSpots.Length > 0)
-                where = planet.Random.Item(freeSpots);
+                where = b.IsBiospheres ? freeSpots[0] : planet.Random.Item(freeSpots);
             return where != null;
         }
 


gkapulis (changes requested):
The inconsistency you found is real — I traced both paths and they do diverge. Double-click goes `OnBuildableItemDoubleClicked` -> `BuildIt` -> `Build(Building)` -> `Enqueue(b, null, true)` -> `AssignBuildingToTile` -> `planet.Random.Item(freeSpots)`, with nothing in that chain aware of terraformable ground, while the governor's `GetPreferredTile` avoids it. Worth closing.

But the justification needs to change before the fix does, because the harm you describe does not happen, and what actually goes wrong points at a slightly different fix.

## A biosphere does not take the tile away from the terraformer

```cs
// PlanetGridSquare.cs:43
public bool CanTerraform => Terraformable && (!Habitable || Habitable && Biosphere);

// Planet_Generate.cs:331
CompleteTileTerraforming(random, TilesList.Filter(t => !t.Habitable && t.Terraformable || t.BioCanTerraform));
```

A terraformable tile with a biosphere on it is **explicitly still a terraforming target** — `BioCanTerraform` is in the candidate filter. And when the terraformer lands there, `MakeTileHabitable` (`Planet_Generate.cs:472`) opens with:

```cs
if (tile.Biosphere)
    ClearBioSpheresFromList(tile);
```

which removes the biosphere from `BuildingList` and drops its upkeep (`Planet_ConstructionQueue.cs:361`). The tile is not blocked and the biosphere is not wasted — it is recycled, and the tile ends up natively habitable with `Terraformable = false`. That is designed behaviour, not an oversight.

## What actually goes wrong

`CompleteTileTerraforming` picks **at random** from that candidate set, and a bio'd tile is already habitable — so terraforming it produces no new habitable space, it just converts and refunds. Every biosphere sitting on terraformable ground therefore dilutes the terraformer's random target pool, spending terraform progress on tiles that are already habitable while genuinely uninhabitable ones wait their turn.

That is a real cost and a fair reason to make the manual path match the governor. It is just a smaller and different claim than "takes the tile the terraformer was waiting for", and the PR description should say so — the next person to read this code will believe whatever the commit message told them.

One more correction while you are in there: the description says `GetPreferredTile` "only falls back to a terraformable one when the colony has nothing else free". It never falls back to a terraformable tile — **both** branches require `!t.Terraformable`. When neither matches it returns `null`, and it is `Enqueue`'s null path that does the random pick. Similar outcome, different mechanism.

## Three changes

**1. `IsBuildingUnlocked(TerraformerId)` is the wrong gate.** Tile terraforming needs terraforming level 2, not merely the unlocked building:

```cs
// Planet_Generate.cs:426, inside Planet.Terraformable
|| terraLevel >= 2 && HasTilesToTerraform
```

`Empire.CanTerraformPlanetTiles` (`Empire.cs:533`) is exactly `IsBuildingUnlocked(TerraformerId) && TerraformingLevel >= 2`, and it is currently declared and never called anywhere — which reads like it was written for this and never wired up. Use it. As the PR stands, an empire that unlocked the Terraformer but sits below level 2 reserves terraformable tiles for terraforming that can never happen.

Worth using it in `GetPreferredTile` too, for the same reason — that one has the same bug, and you are better placed than anyone to fix both at once.

**2. It is not the governor's rule, so please either match it or say so.** `GetPreferredTile` prefers empty tiles first (`!t.BuildingOnTile`) and only then allows an occupied one. Your loop takes the first `!Terraformable` tile in `freeSpots` regardless of occupancy. That difference is defensible — a biosphere does not take the `Building` slot (`PlaceBuilding` just sets `Habitable` and `Biosphere` and returns), so biasing toward already-occupied tiles keeps empty ones free for real buildings, which I would argue is better than what the governor does. But the PR says it "applies the governor's rule", and it does not.

**3. Put the rule in one place.** It now lives in `GetPreferredTile` and in `AssignBuildingToTile`, in two subtly different forms. One helper that both call, with the governor passing `null` and letting the assignment path decide, means they cannot drift — and it pays for itself immediately, because of this:

`GetPreferredTile` filters on `!Habitable`, `!Terraformable` and `!BuildingOnTile`, but never checks `QItem`, volcano or lava. `Enqueue` then validates with `CanEnqueueBuildingHere`, which is `NoQueuedBuildings && !Habitable && !VolcanoHere && !LavaHere`. So when the first empty non-terraformable tile already has something **queued** on it, the governor hands back that tile, validation fails, `Enqueue` returns false, and the biosphere is silently never queued at all — with other perfectly valid tiles standing free. It repeats every evaluation until that tile's building completes. A colony with an active build queue hits this routinely.

That is a bigger bug than the one this PR fixes, it is three lines from your change, and consolidating removes it rather than patching it.

## Fine as-is

The `b.IsBiospheres ? freeSpots[0] : planet.Random.Item(freeSpots)` split is sound — every other building keeps its random pick, and the governor path only reaches it when `GetPreferredTile` already returned null, so the change there is determinism and nothing else. Drag-and-drop onto a specific tile still routes through `Enqueue(b, tile, true)` and only validates, so a player who deliberately wants a biosphere on terraformable ground can still put one there. Good — the fix changes the default without taking away the choice.

Also worth saying: iterating `freeSpots` in list order rather than randomly is the right instinct independent of everything above. Two identical colonies laying out differently because of RNG is a thing players notice and cannot explain.

gkapulis:
Landed as **441802ecb** on `jupiter-1.60/fixes_33`, with a test. Closing this rather than asking for another revision, because the rule it needed to share only came into existence in the two commits before it.

To be straight about the review history: I pushed back on the premise, and I still think that part was wrong. A biosphere on terraformable ground is not lost to the terraformer — `CanTerraform` includes bio'd tiles, `TerraformTiles` filters on `BioCanTerraform`, and `MakeTileHabitable` recycles the biosphere out of `BuildingList` when the tile turns. The real cost is narrower: it dilutes the random target pool that `CompleteTileTerraforming` draws from.

**What you were right about is the part that mattered, and it is what shipped.** There were two rules for the same decision. `BuildableListItem.cs:72` queues a building with no tile, biospheres are in that list until the planet is fully habitable, so a player double-clicking one fell through to `AssignBuildingToTile`'s `Random.Item` while the governor's biosphere followed a considered rule. The player got the worse one, non-deterministically. That divergence is real and it is now closed: the fallback routes biospheres through `Planet.PreferredBiosphereTile`, the same method the governor calls.

Two details differ from your implementation, both because the rule itself changed underneath in **54e182a1f**:

- The test is `Owner.CanTerraformPlanetTiles`, not `IsBuildingUnlocked(TerraformerId)`. This is the correction from the earlier review: turning a tile is terraforming level 2, so between unlocking a terraformer and reaching level 2 there is nothing to save terraformable ground for. Your version would have sent the biosphere past perfectly usable ground for a capability the empire does not have yet.
- Bare ground comes first. `CanPlaceBuildingHere` for a biosphere is only `!Habitable && !VolcanoHere && !LavaHere` — it does not check for a building — so `freeSpots[0]` can be a tile already carrying a build-anywhere building. Roofing one buys the colony no room, which is the whole reason it wanted a biosphere. I made exactly this mistake in my own first draft of the shared rule and the review caught it.

One thing I did not fix, in the interest of not overstating the result: `AssignBuildingToTileOnColonize` and `AssignBuildingToTilePlanetCreation` reach a different picker that special-cases biospheres without considering terraformable ground. Those run at colonization and galaxy generation, so they cannot contradict this at runtime, but they are a third copy of the idea and they are logged.

The test queues through the real path — `Construction.Enqueue(bio, null, playerAdded: true)` — with one acceptable tile among twelve, so a random pick cannot pass by luck. I checked that: with the fix removed it failed eight times out of eight, where an earlier weaker version of the test passed one time in eight.

Thanks for pressing on this one. The premise needed adjusting but the divergence you spotted was real, and it is the kind that quietly stays wrong for years.

gkapulis:
Landed as 441802ecb reusing the shared rule from 54e182a1f; see the previous comment.

